PR #75, opened 2023-01-23
Optimize warp reduction for mapreduce

max-Hawkins:
This uses warp intrinsics to accelerate mapreduce (but doesn't fix a larger underlying issue causing the majority of its slow performance #46). In simple testing, resulted in significant (~2x) speedups for large, non-inplace reductions.

Dependent on #73.


maleadt (approved):
LGTM.

PR #112, opened 2023-03-03
Reduce multiple consecutive values in each thread to improve efficiency

maxwindiff:
Using ideas from:
* https://kieber-emmons.medium.com/optimizing-parallel-reduction-in-metal-for-apple-m1-8e8677b49b01
* https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf

This replaces the old "reduce serially across chunks of input vector that don't fit in a group" loop, which doesn't seem to apply to Metal.jl since we always launch enough threadgroups to cover the entire input.
  
In a simple test this is 2-3x faster than HEAD and achieves > 130GB/s in the 1D reduction case. I'll need to experiment with the ~~stride~~ grain size parameter and document it, and test more complex cases like such as non-power-of-2 sizes or reducing only certain dimensions.

- [x] Experiment with grain size parameter and document it
- [x] Test performance of non-power-of-2 sizes
- [x] Test performance of reducing certain dimensions of N-D arrays
- [x] Check the behavior when number of threadgroups is too high
- [x] Write more tests as needed

Before:

```julia
function init(dims...)
  a = Array{Float32}(undef, dims...)
  for i in 1:length(a)
    a[i] = 1
  end
  return (a, MtlArray(a))
end

a, da = init(8192 * 8192);
b, db = init(8192, 8192);

julia> @btime sum(da)
  6.033 ms (754 allocations: 20.80 KiB)
6.7108864f7

julia> @btime sum(db)
  7.424 ms (759 allocations: 21.33 KiB)
6.7108864f7
```

After:

```julia
julia> @btime sum(da)
  2.030 ms (775 allocations: 21.20 KiB)
6.7108864f7

julia> @btime sum(db)
  3.009 ms (780 allocations: 21.72 KiB)
6.7108864f7
```

Helps with #46

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -211,9 +211,7 @@ function GPUArrays.mapreducedim!(f::F, op::OP, R::WrappedMtlArray{T},
     #
     # also, make sure the grain size is not too high so as to starve threads of work.
     other_groups = length(Rother)
-    while grain > 1 && length(Rreduce) <= reduce_threads * prevpow(2, grain)
-        grain >>= 1
-    end
+    grain = min(grain, prevpow(2, cld(length(Rreduce), reduce_threads)))
     reduce_groups = cld(length(Rreduce), reduce_threads * grain)
 
     # determine the launch configuration
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -91,13 +91,12 @@ Base.@propagate_inbounds _map_getindex(args::Tuple{}, I) = ()
 # product of the two iterators `Rreduce` and `Rother`, where the latter iterator will have
 # singleton entries for the dimensions that should be reduced (and vice versa).
 function partial_mapreduce_device(f, op, neutral, maxthreads, ::Val{Rreduce},
-    ::Val{Rother}, ::Val{Rlen}, shuffle, R, As...) where {Rreduce, Rother, Rlen}
+    ::Val{Rother}, ::Val{Rlen}, ::Val{stride}, shuffle, R, As...) where {Rreduce, Rother, Rlen, stride}
     # decompose the 1D hardware indices into separate ones for reduction (across items
     # and possibly groups if it doesn't fit) and other elements (remaining groups)
     localIdx_reduce = thread_position_in_threadgroup_1d()
-    localDim_reduce = threads_per_threadgroup_1d()
+    localDim_reduce = threads_per_threadgroup_1d() * stride
     groupIdx_reduce, groupIdx_other = fldmod1(threadgroup_position_in_grid_1d(), Rlen)
-    groupDim_reduce = threadgroups_per_grid_1d() ÷ Rlen
 
     # group-based indexing into the values outside of the reduction dimension
     # (that means we can safely synchronize items within this group)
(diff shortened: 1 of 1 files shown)


maxwindiff:
4 seems to be the ideal grain size based on testing in https://gist.github.com/maxwindiff/63b4e8f64c73c4b5b4ff00dd4fb79c5d

It's a bit eerie that the running times are exactly the same for all scalar data types, whether it's `UInt8`, `Float32` or `Int64`, but I don't see anything obviously wrong with the benchmarking code.

maleadt:
Maybe verify execution times with `Metal.@profile` and Xcode?

maxwindiff:
I don't seem to be able to capture timing information. Do you have any idea what I'm missing? I remember seeing the timing graphs once, but I couldn't get it to work again.

```julia
julia> da = MtlArray(fill(Float32(1), 8192 * 8192));
2023-03-04 16:02:35.649 julia18[9346:290846] Metal GPU Frame Capture Enabled

julia> Metal.@profile sum(da)
[ Info: GPU frame capture saved to /Users/kichi/.julia/dev/Metal/julia_capture_1.gputrace/
6.7108864f7
```

Screenshot 2023-03-04 at 16 05 56
Screenshot 2023-03-04 at 16 05 50
Screenshot 2023-03-04 at 16 06 25

maxwindiff:
I think I know what happened, I didn't use `$<var>` in `@btime`. Need to redo the tests.

maxwindiff:
Re-did the benchmarks, this time things look more reasonable: https://gist.github.com/maxwindiff/3ff4886b3e6c1d8ba9ecafb13f84eab7

A grain size of `16 / sizeof(T)` generally works well, except when reducing along certain dimensions of an n-d array. I can see why it isn't helpful to, say, read multiple values along the 2nd dimension of a 3D array. But it was surprising to see reading along the 1st dimension slowing things down... I need to investigate more.

```
sum((256, 512, 512); dims=[1])
grain=1 :  7.610 ms (272 allocations: 7.86 KiB)
grain=2 :  8.056 ms (273 allocations: 7.88 KiB)
grain=4 :  9.150 ms (273 allocations: 7.88 KiB)
grain=8 :  11.315 ms (273 allocations: 7.88 KiB)
grain=16:  20.130 ms (273 allocations: 7.88 KiB)

sum((256, 512, 512); dims=[2])
grain=1 :  9.178 ms (278 allocations: 7.95 KiB)
grain=2 :  8.769 ms (278 allocations: 7.95 KiB)
grain=4 :  9.287 ms (278 allocations: 7.95 KiB)
grain=8 :  12.515 ms (278 allocations: 7.95 KiB)
grain=16:  22.905 ms (278 allocations: 7.95 KiB)
```

maxwindiff:
Made a few more changes:
* Only read multiple values if size of reduction dimension is greater than threads_per_threadgroup (otherwise we'll just starve threads of work for no good reason; this is why `sum((256,512,512); dim=1)` was slow before)
* Only read multiple values if reduction dimension is potentially contiguous in memory

Now performance is better across the board:
HEAD: https://gist.github.com/maxwindiff/4d861ae188c89ea48dd9c2986e349ca4
This PR: https://gist.github.com/maxwindiff/dcb8f3bcc926cbfc3ef55940fd35f345

maxwindiff:
There does't seem to be a limit to the number of threadgroups. For example, this is pretty close to the maximum device buffer size and it runs fine:

```julia
julia> a = MtlArray(fill(1, 30000, 30000, 2));

julia> sum(a; dims=3)
30000×30000×1 MtlArray{Int64, 3}:
[:, :, 1] =
 2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  …  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2
 2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2     2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2
...
```

So I think not having the `ireduce += localDim_reduce * groupDim_reduce` loop is fine.

maxwindiff:
The reduction tests in gpuarrays seems pretty comprehensive already. This is ready for review.

maxwindiff:
Actually thinking about it more, the array dimensions in `gpuarrays/reductions` are all pretty small and would not trigger multi-read. I should add a few test cases with larger array sizes.

maxwindiff:
Adding a test here - https://github.com/JuliaGPU/GPUArrays.jl/pull/459
It passes on nightly, but on other platforms it had a failure when testing `Float32`. I couldn't reproduce the failure locally on 1.8 however. @maleadt do you have any suggestions on how to investigate?

maxwindiff:
Done, rebased! Yep this passes the new GPUArrays tests.

PR #144, opened 2023-03-19
Error when calling device functions on CPU

### LINKED ISSUE #90: Improved errors when calling device functions on CPU
Currently, calling a device function in CPU code causes an ambiguous error and then an infinite hang:

```
julia> thread_position_in_grid_1d()
JIT session error: Symbols not found: [ _julia.air.thread_position_in_grid.i32 ]
JIT session error: Symbols not found: [ _julia.air.thread_position_in_grid.i32 ]
```

I realized this when trying to call `thread_execution_width()`, not realizing it was device-only, but it wasn't clear to me if it wasn't properly wrapped or if something else was going on.

Maybe a clearer error could be raised?

christiangnrd:
Closes #90

### DIFF
--- a/src/device/intrinsics/arguments.jl
+++ b/src/device/intrinsics/arguments.jl
@@ -21,7 +21,7 @@ for (intr, offset) in nodim_intr
     @eval begin
         export $(Symbol(intr))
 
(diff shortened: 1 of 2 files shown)


maleadt (approved):
LGTM, thanks!

PR #182, opened 2023-05-21
Topk for MPSMatrix

christiangnrd:
This was already mostly implemented when I found out that `MPSMatrixFindTopK` only works for `k <= 16`.

Open to feedback on the interface/code organization.


lib/mps/matrix.jl
```diff
@@ -148,3 +148,75 @@ function matmul!(c::MtlMatrix, a::MtlMatrix, b::MtlMatrix,
...
+                                                     where {T<:MtlFloat}
+
+Compute the top `k` values and their corresponding indices column-wise in a matrix `A`.
+Return the indices in `I` and the values in `V`.
+
+`k` cannot be greater than 16.
+
+See also: [`topk`](@ref).
+"""
+function topk!(A::MtlMatrix{T}, I::MtlMatrix{Int32}, V::MtlMatrix{T}, k) where {T<:MtlFloat}
```
tgymnich, line 185: note: The documentations specifies `I` to be of type `MPSDataTypeUInt32`.

lib/mps/matrix.jl
```diff
@@ -148,3 +148,75 @@ function matmul!(c::MtlMatrix, a::MtlMatrix, b::MtlMatrix,
...
+    k <= 16 || error("MPS.topk! does not support values of k > 16")
+    _topk!(A,I,V,k)
+end
+@inline function _topk!(A::MtlMatrix{T}, I::MtlMatrix{Int32}, V::MtlMatrix{T}, k) where {T<:MtlFloat}
+    # Create MPS-compatible matrix from the MtlArrays
+    mps_a = MPSMatrix(A)
+    mps_i = MPSMatrix(I)
+    mps_v = MPSMatrix(V)
+
+    topk_kernel = MPSMatrixFindTopK(current_device(), k)
```
tgymnich, line 195: A few assertions here could catch some common errors:

- Both result matrices must be large enough to hold matrix of size rows x k
- The source matrix must be large enough to contain at least k values

lib/mps/matrix.jl
```diff
@@ -148,3 +148,75 @@ function matmul!(c::MtlMatrix, a::MtlMatrix, b::MtlMatrix,
...
+
+    # Encode and commit topk kernel
+    cmdbuf = MTLCommandBuffer(global_queue(current_device()))
+    encode!(cmdbuf, topk_kernel, mps_a, mps_i, mps_v)
+    commit!(cmdbuf)
+
+    I .+ 1 ,V
```
tgymnich, line 202: The `MPSMatrixFindTopK` kernel has a property to set the `indexOffset` which could avoid starting another kernel here.


```
/*! @property   indexOffset
 *
 *  @discussion Specifies a number that will be added to all the indices written to
 *              resultIndexMatrix in encodeToCommandBuffer. This value can be used
 *              to offset later computations for example by adding the value for
 *              the source matrix column offset sourceMatrixOrigin.y.
 *              Example: Let numberOfTopKValues be 3, let the source be the following:
 *
 *                  source = [ 6.0, 3.0, 8.0, 1.0, 9.0, 4.0, 5.0 ]
 *
 *              and let the sourceMatrixOrigin.y = 2.
 *
 *              Then if indexOffset = 2 then the result value and result index matrices will be:
 *
 *                  result values  = [ 9.0, 8.0, 5.0 ]
 *                  result indices = [  4 ,  2 ,  6  ],
 *
 *              which gives the user indices into the original source matrix.
 *
 *              On the other hand if the indexOffset = 0 then the results  are as follows:
 *
 *                  result values  = [ 9.0, 8.0, 5.0 ]
 *                  result indices = [  2 ,  0 ,  4  ],
 *
 *              which on the other hand gives the user indices into the submatrix starting
 *              from sourceMatrixOrigin.y == 2.
 *
 *              This property is modifiable and defaults to 0. If a different behavior
 *              is desired then this should be modified prior to encoding the kernel.
 */
@property (readwrite, nonatomic) NSUInteger indexOffset;
```

lib/mps/matrix.jl
christiangnrd: If I understand correctly that means I should replace `Int32` with `UInt32`?

lib/mps/matrix.jl
christiangnrd: Thanks! I've implemented the change. 

Btw how do you access that documentation? it's much more detailed than what I found on Apple's website.

tgymnich (approved):

PR #186, opened 2023-05-24
Refactor matmatmul code for faster load time

dkarrasch:
This is in preparation of https://github.com/JuliaGPU/GPUArrays.jl/pull/472. With this one, that PR should pass tests. Afterwards, no modifications are required here, but eventually (when the lower bound is lifted to v1.10), the extra methods can be deleted.

### DIFF
--- a/lib/mps/linalg.jl
+++ b/lib/mps/linalg.jl
@@ -1,5 +1,5 @@
 using LinearAlgebra
-using LinearAlgebra: MulAddMul, wrap
+using LinearAlgebra: MulAddMul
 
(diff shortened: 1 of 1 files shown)


dkarrasch:
Awesome, the Julia nightly test ran on one commit after the threshold. This demonstrates that the "old" branch will eventually become obsolete. Locally, I don't see a slightly trend decreasing trend in package load time, but perhaps this needs to be considered together with the GPUArrays.jl PR.

dkarrasch:
I think this is ready and good to go.

PR #333, opened 2024-04-14
Add MPSMatrixSoftMax

christiangnrd:
And the option to create an MPSMatrix from an MtlVector. Useful since many kernels are only available for MPSMatrix and not MPSVector.

Should be a much quicker review than MPSMatrixRandom ;)

### DIFF
--- a/lib/mps/matrix.jl
+++ b/lib/mps/matrix.jl
@@ -128,6 +128,27 @@ function MPSMatrix(arr::MtlMatrix{T}) where T
     return obj
 end
 
+"""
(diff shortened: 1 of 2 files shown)


christiangnrd:
Test failure is unrelated. (#329)

maleadt (approved):
LGTM. Would be good to add a test, though?

christiangnrd:
> LGTM. Would be good to add a test, though?

I added a few basic tests. I've been wondering how to test for these wrapper functions. Should we just be testing that the objects get constructed properly and leave other tests for higher level APIs?

Issue #336, opened 2024-04-22
KernelAbstractions copyto! typo


101001000:
in MetalKernels.jl:
```
36  function KA.copyto!(::MetalBackend, A::MtlArray{T}, B::MtlArray{T}) where T
37      if device(dest) == device(src)
38          GC.@preserve A B copyto!(A, B)
39          return A
40      else
41          error("Copy between different devices not implemented")
42      end
43  end
```
Calling KA.copyto! with a metal backend will throw: ERROR: LoadError: UndefVarError: `dest` not defined. Maybe dest and src should be A and B respectively?

PR #337, opened 2024-04-22
Typo fix (#336)

101001000:


christiangnrd:
Good catch!
For clarity, I think it would be best to do the opposite and replace `A` with `dest` and `B` with `src`. (For all 3 `copyto!(::MetalBackend, ...)` implementations.)

If that's not something you have the bandwidth to do, let me know!

101001000:
Perfect, I’ll do that

101001000:
It should be done, if you need me to add something more, let me know

tgymnich (approved):

Issue #338, opened 2024-04-26
Segmentation Faults


tknopp:
I get segmentation faults when using Metal on an M1Max
```julia
julia> x = Metal.randn(32)

[54778] signal (11.1): Segmentation fault: 11
in expression starting at REPL[4]:1
objc_msgSend at /usr/lib/libobjc.A.dylib (unknown line)
getproperty at /Users/knopp/.julia/packages/ObjectiveC/JndPi/src/syntax.jl:163 [inlined]
#218 at /Users/knopp/.julia/packages/Metal/q9oGt/src/gpuarrays.jl:71
get! at ./dict.jl:479
unknown function (ip: 0x15f305e0c)
_jl_invoke at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-grannysmith-C07ZQ07RJYVY.0/build/default-grannysmith-C07ZQ07RJYVY-0/julialang/julia-release-1-dot-10/src/gf.c:0 [inlined]
ijl_apply_generic at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-grannysmith-C07ZQ07RJYVY.0/build/default-grannysmith-C07ZQ07RJYVY-0/julialang/julia-release-1-dot-10/src/gf.c:3076
default_rng at /Users/knopp/.julia/packages/Metal/q9oGt/src/gpuarrays.jl:70
gpuarrays_rng at /Users/knopp/.julia/packages/Metal/q9oGt/src/random.jl:3 [inlined]
randn! at /Users/knopp/.julia/packages/Metal/q9oGt/src/random.jl:7 [inlined]
#randn#215 at /Users/knopp/.julia/packages/Metal/q9oGt/src/random.jl:21 [inlined]
randn at /Users/knopp/.julia/packages/Metal/q9oGt/src/random.jl:21
unknown function (ip: 0x15f3022af)
_jl_invoke at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-grannysmith-C07ZQ07RJYVY.0/build/default-grannysmith-C07ZQ07RJYVY-0/julialang/julia-release-1-dot-10/src/gf.c:0 [inlined]
[...]

PR #339, opened 2024-04-26
Add error message for running Metal.jl under Rosetta

tgymnich:
Metal.jl does not seem to work under Rosetta (#338) so add a warning for now until we find a fix.

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -14,7 +14,11 @@ macro sync(code)
 end
 
 function versioninfo(io::IO=stdout)
-    println(io, "macOS $(macos_version()), Darwin $(darwin_version())")
+    print(io, "macOS $(macos_version()), Darwin $(darwin_version())")
+    if process_translated()
+        print(io, ", using Rosetta")
+    end
+    println(io)
     println(io)
 
     println(io, "Toolchain:")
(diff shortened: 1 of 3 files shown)


src/utilities.jl
```diff
@@ -14,7 +14,11 @@ macro sync(code)
 end
 
 function versioninfo(io::IO=stdout)
-    println(io, "macOS $(macos_version()), Darwin $(darwin_version())")
+    print(io, "macOS $(macos_version()), Darwin $(darwin_version())")
+    if process_translated()
+        print(io, ", using Rosetta")
+    end
+    println(io)
```
christiangnrd, line 21: What if you implemented the same mechanism forunsupported platforms as in [CUDA.jl](
https://github.com/JuliaGPU/CUDA.jl/blob/2a3cec926ba48f5bb4efbec5c9242259370963fa/src/initialization.jl#L9-L29)
```suggestion
    @assert functional(true)
    
    println(io, "macOS $(macos_version()), Darwin $(darwin_version())")
```

tgymnich: isn't `functional()` on its way out?

```
    # TODO: make errors here (and in submodules/subpackages like cuBLAS and cuDNN) fatal,
    #       and remove functional(), once people sufficiently use weak dependencies.

```

christiangnrd: In that case this looks good to me.

christiangnrd (approved):

maleadt (approved):
LGTM, thanks.

PR #395, opened 2024-07-19
Add expm1

tgymnich:
Add `expm1`


src/device/intrinsics/math.jl
```diff
@@ -384,3 +384,77 @@ else
...
+    r = v + v
+  end
+
+  return r
+end
+
+@device_override function Base.expm1(a::Float32)
+  r = expm1f_scaled_unchecked(a, 1.0f0)
+  # handle severe overflow and underflow
+  if abs(a - 1.0f0 > 88.0f0)
```
christiangnrd, line 456:
```suggestion
  if abs(a - 1.0f0) > 88.0f0
```

src/device/intrinsics/math.jl
```diff
@@ -384,3 +384,77 @@ else
...
+    r = fma(r, r, -1.0f0)
+  end
+  return r
+end
```
christiangnrd, line 460:
```suggestion
end

```

christiangnrd:
Is it a problem that this doesn't work for `a > ~1.03`?

src/device/intrinsics/math.jl
tgymnich: good catch 😄

src/device/intrinsics/math.jl
```diff
@@ -384,3 +384,77 @@ else
...
+#  DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+#  THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+function expm1f_scaled_unchecked(a::Float32, b::Float32)
+  # exp(a) = 2**i * exp(f); i = rintf (a / log(2))
+  j = fma(1.442695f0, a, 12582912.f0)
+  j = j - 12582912.0f0
+  i = convert(Int32, j)
```
christiangnrd, line 419:
```suggestion
  i = unsafe_trunc(Int32,j)
```
Does this work?

src/device/intrinsics/math.jl
```diff
@@ -384,3 +384,77 @@ else
...
+
+  return r
+end
+
+@device_override function Base.expm1(a::Float32)
+  r = expm1f_scaled_unchecked(a, 1.0f0)
+  # handle severe overflow and underflow
+  if abs(a - 1.0f0) > 88.0f0
+    r = fma(r, r, -1.0f0)
+  end
```
christiangnrd, line 458: Do you know what purpose this serves? I haven't looked into the positive direction, but in the negatives it just seems to make the function be wrong for `a < -87.0f0`

tgymnich: You are right. This seems quite pointless. Here is where i found the snippet in question: https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/kernels/expm1f.h

christiangnrd: Seems like the only difference now between cpu and Metal is `Nan` instead of `Inf` for values > ~88. That probably doesn't really matter does it?

src/device/intrinsics/math.jl
```diff
@@ -384,3 +384,68 @@ else
...
+  if j == 0
+    r = v
+  end
+
+  if j == 1
+    r = v + v
+  end
+
+  return r
+end
```
christiangnrd, line 451:
```suggestion
end

```

christiangnrd:
I think once you fix the indentation to use 4 spaces this will be pretty much ready to merge.

src/device/intrinsics/math.jl
tgymnich: I found the original source: https://forums.developer.nvidia.com/t/a-faster-and-more-accurate-implementation-of-expm1f/48085. Will make some updates...

tgymnich: The MLX implementation skipped `asm ("ex2.approx.ftz.f32 %0,%1;" : "=f"(r) : "f"(a))` for whatever reason. Adding this (`r = 2^a`) makes it work as expected.

christiangnrd: Amazing!

christiangnrd (approved):

PR #409, opened 2024-08-23
Remove redundant calls to mtlconvert

tgymnich:
Follow up to https://github.com/JuliaGPU/CUDA.jl/pull/2472


tgymnich:
This is not yet working because we cannot eagerly `mtlconvert` everything without loosing the toplevel `MtlBuffer` and `MtlPointer` objects

PR #436, opened 2024-09-27
Fix benchmarking CI and benchmark Shared and Private storage modes

christiangnrd:


christiangnrd:
I pushed to the wrong repo. Reopening

PR #446, opened 2024-10-02
simdgroup bfloat intrinsics

christiangnrd:
I can't figure out the simdgroup multiply failures.

What's officially supported:
- Metal stardard library relational ones support bfloat (Table 6.3 in the 4.1 [spec](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf)).
- The four basic arithmetic operators
- simdgroup_matrix
- Tensor stuff
- Nothing else

### DIFF
--- a/lib/mps/MPS.jl
+++ b/lib/mps/MPS.jl
@@ -17,9 +17,9 @@ using ObjectiveC, .Foundation
 
 import GPUArrays
 
-using BFloat16s
+using BFloat16s: BFloat16
 
-const MtlFloat = Union{Float32, Float16}
+const MtlFloat = Union{Float32, Float16, BFloat16}
 
 const MPSShape = NSArray#{NSNumber}
 Base.convert(::Type{MPSShape}, tuple::Union{Vector{<:Integer},NTuple{N, <:Integer}}) where N = NSArray(NSNumber.(collect(tuple)))
--- a/lib/mps/MPS.jl
+++ b/lib/mps/MPS.jl
@@ -19,7 +19,7 @@ import GPUArrays
 
 using BFloat16s: BFloat16
 
-const MtlFloat = Union{Float32, Float16, BFloat16}
+const MtlFloat = Union{Float32, Float16}
 
 const MPSShape = NSArray#{NSNumber}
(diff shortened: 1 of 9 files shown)


christiangnrd:
I guess no point testing this on macOS 13. Will update this to run CI on Sequoia for draft PRs once the MIT runners are back available.

christiangnrd:
Will try again once LLVM 17+ are supported as I suspect intrinsic support for bfloat16 is needed to generate the proper code.

christiangnrd:
> Turns out that software-emulated BFloat16 already worked. I wonder if we should warn users about this, but it's tricky, since our current bad datatype detection operates at the LLVM level (and software-emulated BFLoat16 is simply `i16` there).
> 
> _Originally posted by @maleadt in https://github.com/JuliaGPU/Metal.jl/issues/821#issuecomment-4661755579_

It does seem to be an actual problem for simdgroup matrices.

maleadt:
> It does seem to be an actual problem for simdgroup matrices.

Ah, yes. Anything with intrinsics is not going to work...

christiangnrd:
~Shuffles seem to work but I assume that's because the actual value of the bits being moved around don't matter.~

I was completely wrong. I somehow misread the docs.

maleadt:
It only looks like simdgroup is problematic. Idea: what if we have a GPUCompiler pass detecting the `bf16` intrinsic suffix with `i16` inputs, inserting a bitcast? Going to try this out...

christiangnrd:
Okay I've turned this PR into a smaller PR that's just for simdgroup bfloat (and a drive-by addition to our MPSDataType map)

This still has the issue of only working on 1.13 which I believe 

> It only looks like simdgroup is problematic. Idea: what if we have a GPUCompiler pass detecting the `bf16` intrinsic suffix with `i16` inputs, inserting a bitcast? Going to try this out...

addresses.

Hacky shuffle workarounds can be a separate PR if we so desire

maleadt:
https://github.com/JuliaGPU/GPUCompiler.jl/pull/843 should fix the codegen bits.

maleadt:
Pushed some changes I had locally, which could benefit from some clean-up though, so feel free to rework.

christiangnrd:
API validation test failure is #836

christiangnrd:
@maleadt I think this is ready to go assuming tests don’t fail worse than main

PR #450, opened 2024-10-04
Don't report benchmarks on main branch commits

christiangnrd:
I don't know if this'll work, we'll have to commit and see :)


christiangnrd:
Merging now to see if this fixes it.

PR #483, opened 2024-11-30
Use `OncePerProcess` in 1.12+

christiangnrd:
Ref https://github.com/JuliaLang/julia/pull/55793


maleadt (approved):
LGTM. Might be worth deduplicating the code that determines the value, but probably not worth it here.

Issue #381, opened 2024-07-04
M1/M2: Large matrix multiplications can contains NaNs


chengchingwen:
MWE:

```julia-repl
julia> a = Metal.randn(10000, 10000);

julia> b = Metal.randn(10000, 10000);

julia> c = a * b';

julia> for i in 1:10
           C = Metal.zeros(Float32, size(a))
           mul!(C, a, b')
           @assert C ≈ c "$i"
       end
ERROR: AssertionError: 1
Stacktrace:
 [1] top-level scope
   @ ./REPL[58]:4

julia> for i in 1:10
           C = Metal.zeros(Float32, size(a))
           mul!(C, a, b')
           @assert C ≈ c "$i"
       end
ERROR: AssertionError: 8
Stacktrace:
 [1] top-level scope
   @ ./REPL[58]:4

julia> for i in 1:10
           @assert a * b' ≈ c "$i"
       end
ERROR: AssertionError: 3
Stacktrace:
 [1] top-level scope
   @ ./REPL[59]:2

julia> for i in 1:10
           @assert a * b' ≈ c "$i"
       end
ERROR: AssertionError: 8
Stacktrace:
 [1] top-level scope
   @ ./REPL[59]:2

```

chengchingwen:
adding `wait_completed` on [`matmul!`'s command buffer](https://github.com/JuliaGPU/Metal.jl/blob/abb496d6988a539bbaff7aef205651eccdecec6c/lib/mps/matrix.jl#L235-L237) does not help

christiangnrd:
Adding `Metal.@sync` to the `mul!` also does not help.

maleadt:
I cannot reproduce at all on Metal.jl#master using an M3 Pro, but it does seem reproducible on an M1 Pro.

I wonder if this is a problem with `mapreduce`, since you're calling `isapprox` on GPU arrays. Can you test if calling `@assert Array(C) ≈ Array(c)` makes things pass? It does here, at least.

tgymnich:
I can reproduce the issue on M1. I also looks like all the tasks run on the same queue.

chengchingwen:
The issue was found on a M2 Max. The MWE only happens if the array is large enough. It seems to be launching the subsequent kernel before the matmul finished. Is it possible that the `mapreduce` not checking the availability of the input arrays?

p.s. I'm about to board the plane to JuliaCon so I won't be able to test it soon.

maleadt:
[...]

PR #490, opened 2024-12-05
[dont merge] Check to see effect on performance

christiangnrd:
Pulls in a branch of ObjectiveC.jl that calls `GC.gc(false)` after every autoreleasepool call. With it, I have been unable to reproduce https://github.com/JuliaGPU/Metal.jl/issues/381


christiangnrd:
Ok this is clearly way too hard on the GC to be a reasonable solution.

PR #493, opened 2024-12-10
Fix copy tests

christiangnrd:
MPSDataType enum definition had incorrect Complex numbers definitions. I also forgot that adjoint and transpose  are different for those types when I originally wrote the test.

Only the first and last commit have functional changes


christiangnrd:
I also thought that 15.2 RC2 had fixed MPSCopy with `eltype == ComplexF16` but it must have just been a fluke... Anyway the transpose part should still be fixed.

PR #538, opened 2025-02-07
Use GPUToolbox.jl

christiangnrd:
Should be gtg if tests pass

TODO:
- [x] Remove `[sources]` entry
- [x] Add compat entry


christiangnrd:
I'll merge this one since the only code from GPUToolbox that this uses is `SimpleVersion` and I'm very confident that it's the same code.

PR #566, opened 2025-03-17
Initial MPSGraph support

christiangnrd:
This is ready for review once #565 is merged.

This adds the basic functionality for future extension of MPSGraph framework, and what's needed to replace the MPS matrix multiplication to work around #381.

Support for contiguous views (MtlArrays with an offset) should be added in a future PR, for now it will use the old method

Batched matmul is at the same level of support as the current code, but I think I can use [`broadcastTensor`](https://developer.apple.com/documentation/metalperformanceshadersgraph/mpsgraph/broadcast(_:shape:name:)?language=objc) to enable batched matmul between 1 and N (or N and 1) matrices. Would potentially unblock https://github.com/FluxML/NNlib.jl/pull/614, but that can also be a future PR.


maleadt:
> This is ready for review once #565 is merged.

FYI, you can have PRs target other PRs by setting the base branch, which should make it possible to review (only showing the relevant diff) while auto-updating to `main` when the target PR is merged.

> Close #381

I'd keep that issue open, given that there's MWE's using MPS directly.

What's the performance of MPSGraph like compared to using MPS directly? No difference, I would assume?

christiangnrd:
Very nice about the PR target thing that’ll be very useful for future PRs

I unlinked #381 for automatic closing.

As for performance, I want to make some graphs, but from some quick benchmarks while testing, it seems like it’s slower for smaller matrices, and similar performance for larger ones. That’s why I only made MPSGraphs the default for the types affected by #381.

christiangnrd:
Okay. I wrote a little script to compare MPS and MPSGraph matmul performance. It runs all the supported type variations and saves it to a plot.

<details>
<summary>flopscomp.jl</summary>

```julia
using Pkg
Pkg.activate(temp=true)
Pkg.add(url="https://github.com/christiangnrd/Metal.jl/", rev="MPSGraph")
Pkg.add(["GPUArrays", "Plots"])

# Uncomment if you want to compare with CPU
# Pkg.add(["AppleAccelerate"])
# using AppleAccelerate

using Metal, GPUArrays, LinearAlgebra

using Plots
using Plots.Measures

function cpupeakflops(; n::Integer=4096,
                        n_batch::Integer=1,
                        inT::DataType=Float32,
                        outT::DataType=inT,
                        ntrials::Integer=4,
                        verify=true)
    t = Base.zeros(Float64, ntrials)
    n_batch == 1 || @warn "n_batch > 1 not supported for `mul!`, running with n_batch=1"
    n_batch = 1
    shape = (n, n)
    for i=1:ntrials
        c = zeros(outT, shape...)
        a = ones(inT, shape...)
        b = ones(inT, shape...)
        t[i] = @elapsed mul!(c, a, b)
        verify && @assert only(unique(Array(c))) == n
    end

    return n_batch*2*Float64(n)^3 / minimum(t)
end
function _peakflops(f, n, n_batch, inT, outT, ntrials; verify=true)
    t = Base.zeros(Float64, ntrials)
    shape = n_batch == 1 ? (n, n) : (n, n, n_batch)
    for i=1:ntrials
        c = mtl(zeros(outT, shape...))
        a = mtl(ones(inT, shape...))
        b = mtl(ones(inT, shape...))
        t[i] = @elapsed Metal.@sync f(c, a, b)
        verify && @assert only(unique(Array(c))) == n
    end

    return n_batch*2*Float64(n)^3 / minimum(t)
end
function gpuarrpeakflops(; n::Integer=4096,
                           n_batch::Integer=1,
                           inT::DataType=Float32,
                           outT::DataType=inT,
                           ntrials::Integer=3,
                           verify=true)
    n_batch == 1 || @warn "n_batch > 1 not supported for `GPUArrays.generic_matmatmul!`, running with n_batch=1"
    _peakflops(n, 1, inT, outT, ntrials; verify) do c, a, b
        GPUArrays.generic_matmatmul!(c, LinearAlgebra.wrap(a, 'N'), LinearAlgebra.wrap(b, 'N'), 1, 0)
    end
end
function mpspeakflops(; n::Integer=4096,
                        n_batch::Integer=1,
                        inT::DataType=Float32,
                        outT::DataType=inT,
                        ntrials::Integer=3,
                        verify=true)
    _peakflops(MPS.matmul!, n, n_batch, inT, outT, ntrials; verify)
end
function graphpeakflops(; n::Integer=4096,
                          n_batch::Integer=1,
                          inT::DataType=Float32,
                          outT::DataType=inT,
                          ntrials::Integer=3,
                          verify=true)
    _peakflops(MPSGraphs.graph_matmul!, n, n_batch, inT, outT, ntrials; verify)
end

function compare(Ns, Fs, inT, outT=inT; n_batch=1, ntrials)
    results = Dict()

    for (_, info_str) in Fs
        results[info_str] = Float64[]
    end

    prefixstr = "\33[2K\r($inT, $outT) "
    @time "$((inT, outT))" begin
        for n in Ns
        # for n in Ns
            n_str = "$n: "
            for (f, info_str) in Fs
                print(prefixstr, n_str, info_str)
                push!(results[info_str], f(; inT, outT, n, n_batch, ntrials, verify=(n < maxintfloat(outT))))
                GC.gc()
            end
        end
        print("\33[2K\r")
    end
    return results
end

function main(; #Ns=[50, 64, 100, 128, 250, 256, 500, 512, 1000, 1024, 2000, 2048, 4000, 4096],
                Ns=[50, 64, 100, 128, 250, 256, 500, 512, 1000, 1024, 2000, 2048, 4000, 4096, 8000, 8192],#, 10000],
                Fs=[
                    # (cpupeakflops, "CPU (AppleAccelerate)"), # Uncomment to test CPU performance
                    # (gpuarrpeakflops, "GPUArrays matmatmul"),
                    (mpspeakflops, "MPS matmatmul"),
                    (graphpeakflops, "MPSGraph matmatmul"),
                   ],
                n_batch=1,
                ntrials=5,
                outpath="",
                outtype="svg")
    Ts=[
        (Int8, Float16),
        (Int8, Float32),
        (Int16, Float32),
        (Float16, Float16),
        (Float16, Float32),
        (Float32, Float32),
        ]

    res = Dict()
    for (inT, outT) in Ts
        tmpres = compare(Ns, Fs, inT, outT; n_batch, ntrials)

        plt = plot(xlabel="N, n_batch=$(n_batch)", legendtitle="($inT, $outT)")
        for (res, (_, info_str)) in zip(tmpres,Fs)
            plot!(plt, Ns, tmpres[info_str]; linewidth=1.5, label=info_str)
        end
        res[(inT,outT)] = (plt=plt, results=tmpres)
    end

    finalplot = plot(res[Ts[1]].plt, res[Ts[2]].plt, res[Ts[3]].plt, res[Ts[4]].plt, res[Ts[5]].plt, res[Ts[6]].plt, layout=(2,3),
                     ylim=(0,9e12),
                     plot_title="MatMul Peakflops",
                     tickfonthalign=:left,
                     bottommargin=15pt,
                     size=(2000,1200))
    if !isnothing(outpath)
        savefig(plot(finalplot, dpi=500), joinpath(outpath, "bench_all_$(n_batch).$outtype"))
    end
    return res, finalplot
end

```
</details>

The script activates a temporary environment, installs this branch of Metal, and then compares matmul performance and saves the results to a plot.

To run it, save it to a file, then run:
```julia-repl
julia> include("flopscomp.jl"); res, finalplot = main();
```

I ran it with `GPUArrays.generic_matmatmul!` but that's not needed and makes it take MUCH longer. My results were:
bench_all_1

These results are a bit unfortunate? I don't know if we have a choice though because the combinations where performance is worse are the cases affected by #381...

christiangnrd:
I also ran it on an M1 to get an idea of how long it takes total on the slowest Apple silicon it from installing, precompiling, and running, it took 391 seconds.

bench_all_1_M1


```julia-repl
julia> Metal.versioninfo()
macOS 15.0.1, Darwin 24.0.0

Toolchain:
- Julia: 1.11.4
- LLVM: 16.0.6

Julia packages: 
- Metal.jl: 1.5.1
- GPUArrays: 11.2.2
- GPUCompiler: 1.2.0
- KernelAbstractions: 0.9.34
- ObjectiveC: 3.4.1
- LLVM: 9.2.0
- LLVMDowngrader_jll: 0.6.0+0

1 device:
- Apple M1 (36.086 GiB allocated)
```

christiangnrd:
https://github.com/JuliaGPU/Metal.jl/pull/566/commits/a4e82168a6375e6818857c0d37df4dbb589b7631 makes the operation actually in-place, while https://github.com/JuliaGPU/Metal.jl/pull/566/commits/a86a55d72fdc873ac1daaa058094fda7b65ff259 makes every operation use Float32 under the hood. This makes memory usage quite a bit higher, but greatly improves performance. I'd love opinions on whether we keep it that way or not.

Latest performance on my M2 Max (compare with first screenshot):
bench_all_1_nocopy_float32

maleadt:
> [a86a55d](https://github.com/JuliaGPU/Metal.jl/commit/a86a55d72fdc873ac1daaa058094fda7b65ff259) makes every operation use Float32 under the hood

At the expense of accuracy?

I'm surprised by the large performance differences, but given the issues with MPS matmul we don't really have a choice. That's said, it's good that you kept the `generic_matmatmul!` dispatcher decoupled from the back-end as much as possible, that way we could add some heuristics (small inputs -> MPS, large inputs -> MPS graphs) if we ever figure out the NaN issue.

christiangnrd:
I was trying to figure out why Float16 performance was worse with MPSGraph, and this amazing line added right after initializing the graph gave me some amazing insights.

```julia
function _matmul!(c::MtlArray{Tc}, a::MtlArray{Tab, Na}, b::MtlArray{Tab, Nb}, alpha::Number, beta::Number, transpose_a, transpose_b) where {Tc, Tab, Na, Nb}
    graph = MPSGraph()

    graph.options = MPSGraphs.MPSGraphOptionsVerbose # <- This line

    placeA = placeholderTensor(graph, size(a), Tab)
    placeB = placeholderTensor(graph, size(b), Tab)
...
```

It seems like when the output eltype is `Float16`, MPSGraph runs the operation on the Neural Engine (ANE), which MPS does not do.

Latest commit disables this by default for `graph_matmul!`, restoring performance to what it is when I cast to Float32 without the extra memory usage. This still casts Int8 and Int16 input eltype arrays to the output eltype. **Does that affect precision?** If so, I'll make MPS the default for those types as I was unable to reproduce #381 with Integer input arrays.

New performance:
bench_all_1

I suspect ANE may be better for other M-series processors with weaker GPUs but the same ANE as my 30-core M2 Max. I'll ask on Slack if people want to benchmark on their device and report performance here.

haakon-e:
Results from your script above. I have quite a few programs open (slack, vscode, browsers, etc.) but didn't use anything actively while the test ran.

bench_all_1

sinhtrung:
I ran the code posted on a M2 Macbook Pro
bench_all_1

haakon-e:
Same as above with a MacBook Air:
bench_all_1

christiangnrd:
Thanks @haakon-e @sinhtrung! This is very helpful. I'm very glad to see M4 Pro as I was wondering if the apparently much faster neural engine would make it faster than GPU, but it seems like the TOPS they're referring to may not be Float16 TOPS. (seems like NPUs typically report Int8 OPS)

Seems like we should leave the ANE unused as at best it's as good as the GPU.

christiangnrd:
@AntoineBut I believe I fixed it. LMK if you encounter any other issues.

AntoineBut:
Here is what I get on M3 Pro (MacBook Pro)

bench_all_1

chengchingwen:
Results with M2 Max

bench_all_1

lib/mpsgraphs/tensor.jl
```diff
@@ -0,0 +1,120 @@
...
+end
+function placeholderTensor(graph::MPSGraph, shape::MPSShape, dataType::Type, name = "placeholder tensor")
+    obj = @objc [graph::id{MPSGraph} placeholderWithShape:shape::id{MPSShape}
+                                dataType:dataType::MPSDataType
+                                name:name::id{NSString}]::id{MPSGraphTensor}
+    return MPSGraphTensor(obj)
+end
+
+## MPSGraphTensorData.h
+# @objcwrapper immutable=false MPSGraphTensorData <: MPSGraphObject
```
maleadt, line 28: Why repeat these (and others)? It's expected that they are autogenerated.

dev/flopscomp.jl
```diff
@@ -0,0 +1,177 @@
+# using Pkg
```
maleadt, line 1: I'd put the script in the PR instead; code like that easily bitrots if it isn't covered by CI.

lib/mpsgraphs/tensor.jl
```diff
@@ -0,0 +1,120 @@
...
+
+function MPSGraphTensorData(matrix::MPSMatrix)
+    obj = @objc [MPSGraphTensorData alloc]::id{MPSGraphTensorData}
+    tensor = MPSGraphTensorData(obj)
+    finalizer(release, tensor)
+    @objc [tensor::id{MPSGraphTensorData} initWithMPSMatrix:matrix::id{MPSMatrix}]::id{MPSGraphTensorData}
+    return tensor
+end
+
+# rank must be between 1 and 16 inclusive
```
maleadt, line 66: Make that a precondition throwing an ArgumentError?

lib/mpsgraphs/tensor.jl
```diff
@@ -0,0 +1,120 @@
...
+
+function MPSGraphTensorData(vector::MPSVector)
+    obj = @objc [MPSGraphTensorData alloc]::id{MPSGraphTensorData}
+    tensor = MPSGraphTensorData(obj)
+    finalizer(release, tensor)
+    @objc [tensor::id{MPSGraphTensorData} initWithMPSVector:vector::id{MPSVector}]::id{MPSGraphTensorData}
+    return tensor
+end
+
+# rank must be between 1 and 16 inclusive
```
maleadt, line 84: Same.

maleadt (approved):
Great stuff!

Couple of minor comments, but LGTM otherwise.

lib/mpsgraphs/tensor.jl
christiangnrd: Mostly leftover from the transition to automatically generated wrappers as this branch has been chilling on my computer for a few months and it was hard to navigate without these comments.

In some future PR I'll go through all 3 frameworks and remove all the comments where the object has a constructor

lib/mpsgraphs/tensor.jl
christiangnrd: Good idea. I also fixed it for `MPSNDArrayDescriptor`

dev/flopscomp.jl
christiangnrd: Done

tgymnich (approved):

lib/mpsgraphs/tensor.jl
christiangnrd: Ok I was going to start cleaning up the wrappers, but having the type and its supertype there without having to search though the lib*.jl file when wrapping related functions is **very** convenient so I think we should leave them in until everything has been wrapped or I/someone else get around to wrapping the methods.

christiangnrd:
If no one objects, I'll merge this tomorrow morning. 

I'm wondering if we should use MPS instead of MPSGraph in situations where it's faster and not broken. I'll open a PR for feedback soon.

PR #576, opened 2025-03-27
Test 1.12-nightly

christiangnrd:
Suggested here https://github.com/JuliaGPU/Metal.jl/pull/575#issuecomment-2758363545.

I went with replacing nightly for the moment as llvmdowngrader is not yet supported for llvm 18+ so we know CI will fail for both.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -30,10 +30,10 @@ steps:
             julia:
               - "1.10"
               - "1.11"
-              - "nightly"
+              - "1.12-nightly"
           adjustments:
             - with:
-                julia: "nightly"
+                julia: "1.12-nightly"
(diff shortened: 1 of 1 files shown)


maleadt:
Not sure this is valuable without the bumped downgrader; now you're just removing opaque pointer testing from the CI roster without a replacement that will ever pass.

maleadt (changes requested):
Not sure this is valuable without the bumped downgrader; now you're just removing opaque pointer testing from the CI roster without a replacement that will ever pass.

So I wouldn't merge this until we have that JLL at the least.

maleadt:
FWIW, simply rebasing this PR is not going to help. It needs somebody to go in and look at the specific failures, which I currently do not have the time for.

PR #604, opened 2025-06-10
Add bare minimum for macOS 26 Tahoe

christiangnrd:
The errors this fixes only show up when testing with a version of Julia compiled on macOS 26.

I'm thinking of waiting a couple beta versions before adding the rest of the wrappers in case things change.


maleadt (approved):

christiangnrd:
I'll bump patch release with the other little fixes.

PR #614, opened 2025-06-25
More accumulation and reduction benchmarks

christiangnrd:
Adds reduction and accumulation benchmarks for Int64 ~(maybe should be Int32?)~ and for `dims=2` per @anicusan's mention of finding different performance profiles between datatypes and dimensions.

(Do not merge)

### DIFF
--- a/perf/array.jl
+++ b/perf/array.jl
@@ -3,150 +3,156 @@ const n = 1000
 const m_long = 3
 const n_long = 1_000_000
 
+group = addgroup!(SUITE, "array")
+
 for (S, smname) in [(Metal.PrivateStorage,"private"), (Metal.SharedStorage,"shared")]
-    local group = addgroup!(SUITE, "$smname array")
+    let group = addgroup!(group, smname)
 
-    # generate some arrays
-    cpu_mat = rand(rng, Float32, m, n)
-    gpu_mat = MtlMatrix{Float32,S}(cpu_mat)
-    gpu_mat_long = MtlMatrix{Float32,S}(rand(rng, Float32, m_long, n_long))
-    gpu_vec = reshape(gpu_mat, length(gpu_mat))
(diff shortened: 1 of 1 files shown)


christiangnrd:
<details>
<summary>Benchmark results with https://github.com/JuliaGPU/AcceleratedKernels.jl/pull/47 applied:</summary>

image
image
image
</details>

AK is AcceleratedKernels with default settings, and AK 1024 is with `block_size=1024`. The latter is mostly hypothetical as setting it to that valueby default would cause failures when Metal decides it doesn't have that many threads to give a block. However, 
`block_size=256` seems to be the better option for all benchmarks except for one `size` and `dims` configuration (shown in the figures).

In accumulate, we see either similar or (sometimes much) worse performance with AK. The only exception is when `dims=1` with the 3x1000000 Float32 matrix.

Similar story for the reductions, where it seems like AK is much faster for `dims=1` in the 3x1000000 matrices, but the same or worse for the rest.

This benchmarking also made me notice that AK's reduce operations don't support multiple `dims` being specified.

In CUDA's case, the current algorithms seem to be faster in all reductions, and all accumulations except for `dims=2` for the 512x1000 matrices, and `dims=1` for the 3x1000000 matrices.

christiangnrd:
@maleadt This is ready for review. I recommend doing so per commit since the first one (https://github.com/JuliaGPU/Metal.jl/pull/614/commits/bbadbf35c5a66b549761ab78bc49c0d135494235) includes the changes that will be mirrored in the CUDA benchmark PR, while the third one (https://github.com/JuliaGPU/Metal.jl/pull/614/commits/82abf223ff230621edbc0d12b201611085216ec8) is Metal.jl-specific where I split up the tests that are not affected by storage mode so we're not wasting time running redundant benchmarks.

maleadt (approved):

PR #626, opened 2025-07-20
Add function to retrieve # of gpu cores in system

christiangnrd:
This could enable the same heuristic used in CUDA.jl to decide between reduction by block/group or reduction by thread


tgymnich (approved):

maleadt (approved):

christiangnrd:
Letting tests run here would be a watse of CI I’m just going to merge.

PR #633, opened 2025-07-25
Synchronize using `MTLSharedEvents`

christiangnrd:
By reusing the same shared event, reduces the number of allocations.

Hopefully slightly improves performance, but this is more in line with how I believe queue synchronization will have to work with Metal 4.


maleadt (approved):
LGTM.

Is there a Metal3->4 migration document anywhere indicating we should switch from `MTLCommandBuffer::waitUntilCompleted` to a shared event?

christiangnrd:
> all other operations seem slightly faster

Good point. I'll rerun and if results are similar I'll merge

PR #651, opened 2025-08-07
Re-add old valid benchmarks and remove problematic trailing commas

christiangnrd:
Unless suggested otherwise, this is my only attempt at fixing this.

The issue was that browsers are forgiving to trailing commas, so things looked right when I Attempted to verify, but the js function `JSON.parse` that's used to check the json is not, so results were completely overwritten.

### DIFF
--- a/bench/data.js
+++ b/bench/data.js
@@ -1,8 +1,15398 @@
 window.BENCHMARK_DATA = {
-  "lastUpdate": 1754424592449,
+  "lastUpdate": 1754097557293,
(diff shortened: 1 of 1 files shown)


bench/data.js
```diff
@@ -1,8 +1,15398 @@
 window.BENCHMARK_DATA = {
-  "lastUpdate": 1754424592449,
+  "lastUpdate": 1754097557293,
```
christiangnrd, line 2:
```suggestion
  "lastUpdate": 1754424592449,
```

maleadt (approved):

PR #658, opened 2025-09-10
Don't transpose empty MtlArrays

### LINKED ISSUE #656: Metal.jl doesn't allow creation of arrays with size 0 along an axis
See https://buildkite.com/julialang/gpuarrays-dot-jl/builds/1273#01990a94-333f-4dda-a6ee-1ece6e8610bd and https://github.com/JuliaGPU/GPUArrays.jl/pull/613#issuecomment-3245013249

christiangnrd:
Fix #656

Should I backport this to 1.7? `main` has some feature removals so it should probably be 1.8 or 2.0 and I would also like to remove macOS 13 guards and support with the next release.


kshyatt (commented):
Can we add a test?

PR #660, opened 2025-09-11
Switch to `macos-15-intel` for x64 load tests

christiangnrd:
Also make architecture detection in the test script a bit more robust.

This was originally to move to buildkite but then github actions `macos-15-intel` runners were [announced](https://github.com/actions/runner-images/issues/13045).


christiangnrd:
Just found that there are macos 15 intel images. Probably better to use that instead. https://github.com/actions/runner-images?tab=readme-ov-file#available-images

PR #666, opened 2025-10-06
Bump `BFloat16s` compat

christiangnrd:
BFloat16s.jl v0.5.1 will continue to be used until https://github.com/JuliaLLVM/LLVM.jl/pull/526 is merged and registered


christiangnrd:
Let's keep #667 open until merge so it doesn't keep reopening

PR #681, opened 2025-10-17
Add some precompile statements

christiangnrd:
These seem like the most important statements to include for reducing TTFX across things like compiling a kernel and constructing MtlArrays


maleadt:
I take it a precompile workload still crashes because of cache corruption?

PR #682, opened 2025-10-17
Minor CI cleanup and force 3 workers on 8GB mac minis

christiangnrd:
Add storage back to Special tests since it doesn't make sense to have 2 1-job groups.

I'm temporarily printing memory info because I'm trying to figure out how to best handle picking worker count on macs so that we don't end up waiting 37 minutes for a 15-minute job. Will be reverted before merging

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,6 +1,15 @@
 using Metal
 using ParallelTestRunner
 
+# Force 3 workers if running on the 8GB buildkite mac minis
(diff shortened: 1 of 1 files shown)


maleadt:
> I'm temporarily printing memory info because I'm trying to figure out how to best handle picking worker count on macs so that we don't end up waiting 37 minutes for a 15-minute job.

I'd be fine forcing that from the CI pipelines for now until we figure this out in ParalleTestWorker.jl

PR #686, opened 2025-10-20
Add `MTLResidencySet`

christiangnrd:


maleadt (approved):
Any intended use case?

christiangnrd:
`MTLResidencySet` seems to be a Metal 3.2 feature that is required in Metal 4 so this could make that eventual PR smaller.

PR #687, opened 2025-10-21
`MTLHeap` for argument encoding

christiangnrd:
I think this is significantly slower for compute cases but I'm curious to see the effects.
Tests pass locally so I just want the benchmarks.

[only benchmarks]


christiangnrd:
Memory leak somewhere

PR #705, opened 2025-11-23
Add host RNG based on the native kernel RNG

### LINKED ISSUE #691: Replace GPUArrays-based RNG with native one
Basically porting https://github.com/JuliaGPU/CUDA.jl/blob/b0b484c5bf7e7a8342e0285e42c5e235aa252c32/src/random.jl#L12-L190 over

Ideally we move this code over to GPUArrays.jl, but that would mean requiring back-ends to implement device-side `rand()`, which is a big ask.

christiangnrd:
Ported from CUDA.jl
Close #691

### DIFF
--- a/test/random.jl
+++ b/test/random.jl
@@ -258,7 +258,7 @@ end
         randn!(rng, A)
 
         B = Array{T}(undef, dims)
-        Metal.@allowscalar rand!(rng, B)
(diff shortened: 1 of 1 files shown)


maleadt (approved):
LGTM!

PR #717, opened 2025-12-04
Change default storage mode to SharedStorage on UMA devices

KaanKesginLW:
## Summary

Change Metal.jl's default storage mode from `PrivateStorage` to `SharedStorage` on Apple Silicon (unified memory architecture) devices. Intel Macs with discrete GPUs retain `PrivateStorage` as default.

This change:
- Aligns with Apple's official guidance for unified memory platforms
- Enables zero-copy CPU access via `unsafe_wrap(Array, gpu_arr)` (1.8x - 6x faster)
- Has no performance regression (verified with 67 benchmarks)

### 1. Apple's Guidance for Unified Memory

Apple's [Metal Best Practices Guide](https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/MTLBestPracticesGuide/ResourceOptions.html) explicitly states:

> **iOS/tvOS (unified memory)**: *"The Shared mode is usually the correct choice"*

Apple Silicon Macs have the same unified memory architecture as iOS/tvOS. The current `PrivateStorage` default was based on guidance intended for **discrete GPU systems** where CPU and GPU have separate memory pools.

### 2. Performance is Equivalent (67 Tests with Statistical Significance)

Comprehensive benchmarks on M2 Max with 95% confidence interval testing show **no performance difference** at production sizes:

| Metric | Count |
|--------|-------|
| Total tests | 67 |
| Ties (no significant difference) | **52 (78%)** |
| SharedStorage wins | 7 (10%) |
| PrivateStorage wins | 8 (12%) |

**Key findings:**
- At sizes ≥512 MB: **virtually all operations are ties**
- `copyto!`, `fill!`: **Identical** at all sizes
- MPS matmul: **Identical** at all sizes (4/4 ties)
- Element-wise at ≥1 GB: **All ties**

<details>
<summary>Benchmark breakdown by size</summary>

| Size | Private Wins | Shared Wins | Ties |
|------|-------------|-------------|------|
| 64 MB | 1 | 2 | 6 |
| 128 MB | 2 | 2 | 5 |
| 256 MB | 0 | 2 | 7 |
| **512 MB** | **0** | **0** | **10** |
| 1024 MB | 3* | 0 | 7 |
| **2048 MB** | **0** | **1** | **9** |
| **4096 MB** | **0** | **0** | **10** |

*1024 MB "regressions" disappeared with more iterations (measurement noise)
</details>

### 3. SharedStorage Enables Faster CPU Access

SharedStorage enables `unsafe_wrap(Array, gpu_arr)` for zero-copy CPU access, avoiding the allocation + copy overhead of `Array()`:

| Size | Private: `Array()` + use | Shared: `unsafe_wrap()` + use | Speedup |
|------|--------------------------|-------------------------------|--------:|
| 512 MB | 17.0 ms | 9.0 ms | **1.9x** |
| 1 GB | 33.8 ms | 18.1 ms | **1.9x** |
| 2 GB | 65.1 ms | 36.4 ms | **1.8x** |
| 4 GB | 442.5 ms | 73.9 ms | **6.0x** |
| 8 GB | 888.6 ms | 147.3 ms | **6.0x** |

*Benchmarked on M2 Max, median of 5 runs. Both paths include `sum()` to measure actual data access.*

The speedup comes from avoiding allocation and memory copy - with SharedStorage, CPU and GPU share the same physical memory on Apple Silicon.

## Implementation

The default preference changes from `"private"` to `"auto"`, which detects UMA at module load:

```julia
const DefaultStorageMode = let str = @load_preference("default_storage", "auto")
    if str == "private"
        PrivateStorage
    elseif str == "shared"
        SharedStorage
    elseif str == "auto"
        if Sys.isapple() && !isempty(devices())
            MTLDevice(1).hasUnifiedMemory ? SharedStorage : PrivateStorage
        else
            PrivateStorage
        end
    else
        error("unknown default storage mode: $str")
    end
end
```

**Preference options:**
- `"auto"` (new default): SharedStorage on UMA, PrivateStorage otherwise
- `"shared"`: Force SharedStorage
- `"private"`: Force PrivateStorage (previous behavior)

## Backward Compatibility

1. **Preference override**: Users can set `default_storage = "private"` in `LocalPreferences.toml`
2. **Explicit storage**: `MtlArray{T,N,PrivateStorage}(...)` always works
3. **Test suite ready**: Already handles non-Private default (see `test/runtests.jl:85-88`)

## Related

- PR #126: Original switch to PrivateStorage (2022)
- Issue #86: Improve use of unified memory
- PR #716: Fix SharedStorage copyto! performance (3.5x speedup)

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -31,7 +31,7 @@ function check_eltype(T)
     Base.isbitsunion(T) && error("MtlArray does not yet support isbits-union arrays")
     contains_eltype(T, Float64) && error("Metal does not support Float64 values, try using Float32 instead")
     contains_eltype(T, Int128) && error("Metal does not support Int128 values, try using Int64 instead")
-    contains_eltype(T, UInt128) && error("Metal does not support UInt128 values, try using UInt64 instead")
+    return contains_eltype(T, UInt128) && error("Metal does not support UInt128 values, try using UInt64 instead")
 end
 
 """
@@ -43,14 +43,14 @@ end
 
 See the Array Programming section of the Metal.jl docs for more details.
 """
-mutable struct MtlArray{T,N,S} <: AbstractGPUArray{T,N}
+mutable struct MtlArray{T, N, S} <: AbstractGPUArray{T, N}
     data::DataRef{<:MTLBuffer}
 
     maxsize::Int  # maximum data size in bytes; excluding any selector bytes
     offset::Int   # offset of the data in the buffer, in number of elements
     dims::Dims{N}
 
-    function MtlArray{T,N,S}(::UndefInitializer, dims::Dims{N}) where {T,N,S}
+    function MtlArray{T, N, S}(::UndefInitializer, dims::Dims{N}) where {T, N, S}
         check_eltype(T)
         maxsize = prod(dims) * sizeof(T)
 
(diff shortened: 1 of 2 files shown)


christiangnrd:
This will probably eventually happen but it'll be easier to do it in a separate PR than to rebase and clean up this one.

PR #721, opened 2025-12-05
Add async background warmup to reduce first-kernel latency

KaanKesginLW:
## Summary

The first GPU kernel in a Metal.jl session takes ~1.75 seconds due to one-time JIT compilation of the GPU compilation pipeline (GPUCompiler, LLVM passes, etc.). This PR introduces async background warmup during package initialization to reduce this to 0.035-0.20 seconds—a **9-50x improvement** in perceived first-kernel latency.

## Problem

Users experience a jarring ~2 second delay on their first GPU operation:

```julia
using Metal
a = MtlArray(rand(Float32, 1024, 1024))
@time fill!(a, 1.0f0)  # 1.75s - unexpected!
@time fill!(a, 2.0f0)  # 0.001s - fast as expected
```

This causes:
- Misleading benchmark results (first iteration 50x slower)
- Poor first impressions for new users evaluating Metal.jl
- Confusion ("is this a memory issue? a bug?")

### Root Cause Analysis

The delay was previously attributed to memory page faults on large arrays. Investigation revealed this is incorrect—the actual cause is JIT compilation:

| Evidence | Finding |
|----------|---------|
| 1KB array | Same 1.75s delay as 512MB |
| Storage mode | No difference (Private vs Shared) |
| Compilation stages | check_method (0.2s) + LLVM IR gen (1.1s) + AIR (0.1s) |

## Solution

Start a minimal kernel compilation in the background during `__init__()` **when multiple threads are available**. By the time users run their first kernel, most or all initialization is complete.

### Key Discovery

Concurrent compilations share the one-time initialization overhead:

```
Warmup kernel:  1.620s
User kernel:    0.196s  (concurrent, not 1.7s!)
Total wall:     1.808s
```

The user kernel benefits from shared initialization even when warmup hasn't completed, due to lock serialization in `mtlfunction`.

### New Files
- `src/warmup.jl` - Warmup kernel and public `Metal.warmup()` API
- `test/warmup.jl` - Unit tests for warmup functionality

### Modified Files
- `src/initialization.jl` - Add warmup task startup in `__init__()`
- `src/Metal.jl` - Include warmup module

### API Additions

```julia
Metal.warmup(; blocking=true)  # Wait for warmup to complete
Metal.warmup(blocking=false)   # Return immediately
```

Note: `warmup` is not exported to avoid namespace pollution. Call via `Metal.warmup()`.

### Preferences

Users can disable warmup via `LocalPreferences.toml`:

```toml
[Metal]
warmup = false
```

## Performance

| Scenario | Before | After | Improvement |
|----------|--------|-------|-------------|
| Explicit wait | 1.75s | 0.035s | 50x |
| Immediate (concurrent) | 1.75s | 0.20s | 9x |
| Typical workflow | 1.75s | 0.04-0.15s | 12-44x |

## Trade-offs

**What does the user lose?** Nothing meaningful:

| Concern | Impact |
|---------|--------|
| Import time | Unchanged (~1.1s) - warmup runs in background, doesn't block |
| Memory | 4 bytes temporarily allocated, freed immediately |
| CPU | ~1.7s of single-threaded background work |
| Correctness | Unaffected |
| API | No breaking changes |

The background CPU usage is practically unnoticeable on modern Apple Silicon Macs (8+ cores). Benchmarks show <2% overhead on concurrent CPU workloads—well within measurement noise. The compilation work would happen anyway on the user's first kernel; we're simply shifting it to run earlier in the background while the user's code is still setting up.

Users who need to measure cold-start compilation (debugging/profiling) can disable via preference.

### Misleading Benchmarks Lead to Wasted Debugging Time

Without warmup, users comparing CPU vs GPU performance get dramatically wrong conclusions:

```
Matrix multiply (4096×4096 Float32):
  CPU: 0.306s
  GPU (first call):  1.012s  ← User thinks GPU is 3x SLOWER
  GPU (second call): 0.019s  ← Actual: GPU is 16x FASTER
```

A user unaware of this one-time JIT cost might:
- Conclude Metal.jl is slower than CPU and abandon it
- Spend hours debugging a non-existent "performance bug"
- File issues about inconsistent profiling results
- Distrust their own benchmarks

### First Impressions for New Users

(Highly relevant for computational scientists with specializations in biology, neuroscience, chemistry, etc. who might not know or care about compilation mechanics despite being the target audience for Julia)

When someone evaluates Metal.jl for the first time:

```julia
julia> using Metal
julia> a = MtlArray([1, 2, 3])
julia> @time a .+ 1   # 1.7s delay - "is this broken?"
```

This 2-second hang on a trivial operation creates a poor first impression, especially compared to frameworks like PyTorch or CUDA.jl where GPU operations feel instant. With async warmup, the experience becomes what users expect—responsive from the first interaction.

## Testing

All existing tests pass. New tests added:
- Warmup task starts and completes without error
- `Metal.warmup()` API works correctly
- Kernel compilation is fast after warmup
- Concurrent compilations don't deadlock

### Single-threaded REPL blocking

**Concern**: In single-threaded mode, Julia's cooperative scheduling means JIT compilation doesn't yield, potentially blocking the REPL during warmup.

**Response**: Metal.jl users are pursuing GPU computing on Apple Silicon. It's reasonable to expect they've explored CPU parallelism first (setting `-t auto` or `JULIA_NUM_THREADS`), which is typically a prerequisite step before GPU for real end users in scientific computing work.

**Default to old behaviour**: Warmup only runs when `Threads.nthreads() > 1` (i.e., when Julia is started with `-t auto` or `JULIA_NUM_THREADS > 1`).

With a single thread, Julia's cooperative task runtime means an async task would block the main thread during JIT compilation, potentially hurting perceived REPL latency. To avoid this, Metal.jl warmup is skipped entirely in single-threaded mode—users get the same behaviour as before this PR (assuming this helps with perceived responsiveness for these niche users).

This addresses [@vchuravy](https://github.com/vchuravy)'s concern about REPL blocking while still providing the optimization for the common case (multi-threaded Julia for Metal.jl users on Apple Silicon).

### DIFF
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -3,14 +3,14 @@
         try
             dev = device()
             return supports_family(dev, MTL.MTLGPUFamilyApple7) &&
-            supports_family(dev, MTL.MTLGPUFamilyMetal3)
+                supports_family(dev, MTL.MTLGPUFamilyMetal3)
         catch
             return false
         end
     end
 else
     # Becomes `nothing` once it has been determined that the device is on macOS
-    const _functional = Ref{Union{Nothing,Bool}}(false)
+    const _functional = Ref{Union{Nothing, Bool}}(false)
 
     function functional()
         if isnothing(_functional[])
@@ -67,7 +67,7 @@ function __init__()
             _functional[] = nothing  # VERSION <= v"1.12.0-DEV.1421"
         end
     catch err
-        @error "Failed to load Metal" exception=(err,catch_backtrace())
+        @error "Failed to load Metal" exception = (err, catch_backtrace())
         return
     end
 
@@ -78,13 +78,13 @@ function __init__()
     end
 
     # start async warmup to reduce first-kernel JIT compilation latency
-    if functional() && _warmup_enabled
+    return if functional() && _warmup_enabled
         _warmup_task[] = errormonitor(@async _warmup_compilation())
     end
 end
 
 function synchronize_metal_tasks(ex)
-    quote
+    return quote
         try
             $(ex)
         finally
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -24,6 +24,10 @@ else
     end
 end
 
+# Async warmup system to reduce first-kernel JIT compilation latency
+const _warmup_task = Ref{Union{Nothing, Task}}(nothing)
+const _warmup_enabled = @load_preference("warmup", true)
+
 function __init__()
     precompiling = ccall(:jl_generating_output, Cint, ()) != 0
     precompiling && return
@@ -71,6 +75,11 @@ function __init__()
     # because displaying values happens on a different task
     if isdefined(Base, :active_repl_backend) && !isnothing(Base.active_repl_backend)
         push!(Base.active_repl_backend.ast_transforms, synchronize_metal_tasks)
+    end
+
+    # start async warmup to reduce first-kernel JIT compilation latency
+    if functional() && _warmup_enabled
+        _warmup_task[] = errormonitor(@async _warmup_compilation())
     end
 end
 
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -81,7 +81,7 @@ function __init__()
     # Only run with multiple threads - with a single thread, the async task would
     # block the main thread due to Julia's cooperative task runtime.
     return if functional() && _warmup_enabled && Threads.nthreads() > 1
-        _warmup_task[] = errormonitor(@async _warmup_compilation())
+        _warmup_task[] = errormonitor(Threads.@spawn _warmup_compilation())
     end
 end
 
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -77,8 +77,10 @@ function __init__()
         push!(Base.active_repl_backend.ast_transforms, synchronize_metal_tasks)
     end
 
-    # start async warmup to reduce first-kernel JIT compilation latency
-    return if functional() && _warmup_enabled
+    # Start async warmup to reduce first-kernel JIT compilation latency.
+    # Only run with multiple threads - with a single thread, the async task would
+    # block the main thread due to Julia's cooperative task runtime.
+    return if functional() && _warmup_enabled && Threads.nthreads() > 1
         _warmup_task[] = errormonitor(@async _warmup_compilation())
     end
 end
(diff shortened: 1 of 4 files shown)


vchuravy:
Julia uses a cooperative task runtime, so saying that something is `async` doesn't mean that much. If you launch single threaded, the thread will be blocked.

KaanKesginLW:
I can say that this definitely has affected me as an end user, in a way it doesn't anymore. And looking back at my commits from the past few days I can tell that this package was never really tested by a real end user, with all due respect.

src/warmup.jl
```diff
@@ -0,0 +1,71 @@
+# Async warmup to reduce first-kernel JIT compilation latency
+#
+# The first GPU kernel in a Metal.jl session takes ~1.75s due to one-time JIT
+# compilation of GPUCompiler internals. By starting a minimal kernel compilation
+# in the background during __init__(), we can reduce this to 0.035-0.20s for the
+# user's first actual kernel—a 9-50x improvement.
+
+export warmup
+
```
christiangnrd: This should fix the benchmark error.
```suggestion
```

src/initialization.jl
```diff
@@ -72,10 +76,17 @@ function __init__()
     if isdefined(Base, :active_repl_backend) && !isnothing(Base.active_repl_backend)
         push!(Base.active_repl_backend.ast_transforms, synchronize_metal_tasks)
     end
+
+    # Start async warmup to reduce first-kernel JIT compilation latency.
+    # Only run with multiple threads - with a single thread, the async task would
+    # block the main thread due to Julia's cooperative task runtime.
+    return if functional() && _warmup_enabled && Threads.nthreads() > 1
+        _warmup_task[] = errormonitor(@async _warmup_compilation())
```
vchuravy:
```suggestion
        _warmup_task[] = errormonitor(Threads.@spawn _warmup_compilation())
```

`@async` is pinned to the same thread as parent.

maleadt:
This is not a good way to do this. Also has been mostly fixed by recent (and impending) TTFX work:
```julia-repl
julia> using Metal

julia> a = MtlArray([1, 2, 3])
3-element MtlVector{Int64, Metal.PrivateStorage}:
 1
 2
 3

julia> @time a .+ 1
  0.213803 seconds (177.98 k allocations: 8.546 MiB, 52.78% compilation time: 23% of which was recompilation)
3-element MtlVector{Int64, Metal.PrivateStorage}:
 2
 3
 4

julia> @time a .+ 1
  0.000271 seconds (53 allocations: 1.641 KiB)
```

PR #738, opened 2026-01-31
Support division between two ComplexF32 numbers

albertomercurio:
Fixes #736


oscardssmith:
I was referring to https://github.com/JuliaLang/julia/blob/2cf16b10e40956671dde660de2e7914037e8b078/base/complex.jl#L390, but it might be the case that the performance hit is too large.

oscardssmith:
the MLX version is the trivial one. It will suffer from overflow and underflow, but that's true of any algorithm that doesn't do some form of rescaling.

christiangnrd:
Superceeded by #762

PR #747, opened 2026-02-25
Improve time-to-first-kernel

maleadt:
... by adding a proper precompilation workload and removing some overzealous specialization.
Needs to be properly validated to ensure the `@nospecialize` on crucial functions like `mtlfunction` doesn't regress launch overhead.

Before:

```
❯ hyperfine --warmup=1 'julia --project -e "using Metal; @metal identity(nothing)"'
Benchmark 1: julia --project -e "using Metal; @metal identity(nothing)"
  Time (mean ± σ):      2.695 s ±  0.135 s    [User: 3.360 s, System: 0.401 s]
  Range (min … max):    2.588 s …  3.063 s    10 runs

❯ hyperfine --warmup=1 'julia --project examples/vadd.jl'
Benchmark 1: julia --project examples/vadd.jl
  Time (mean ± σ):      3.827 s ±  0.237 s    [User: 4.471 s, System: 0.380 s]
  Range (min … max):    3.573 s …  4.260 s    10 runs
```

After:

```
❯ hyperfine --warmup=1 'julia --project -e "using Metal; @metal identity(nothing)"'
Benchmark 1: julia --project -e "using Metal; @metal identity(nothing)"
  Time (mean ± σ):      1.555 s ±  0.026 s    [User: 2.232 s, System: 0.405 s]
  Range (min … max):    1.516 s …  1.599 s    10 runs

❯ hyperfine --warmup=1 'julia --project examples/vadd.jl'
Benchmark 1: julia --project examples/vadd.jl
  Time (mean ± σ):      2.065 s ±  0.056 s    [User: 2.741 s, System: 0.293 s]
  Range (min … max):    2.007 s …  2.168 s    10 runs
```

### DIFF
--- a/src/compiler/execution.jl
+++ b/src/compiler/execution.jl
@@ -179,11 +179,12 @@ The output of this function is automatically cached, i.e. you can simply call `m
 in a hot path without degrading performance. New code will be generated automatically when
 the function changes, or when different types or keyword arguments are provided.
 """
-function mtlfunction(f::F, tt::TT=Tuple{}; name=nothing, kwargs...) where {F,TT}
+function mtlfunction(@nospecialize(f), @nospecialize(tt)=Tuple{}; name=nothing, kwargs...)
     dev = device()
     Base.@lock mtlfunction_lock begin
         # compile the function
         cache = compiler_cache(dev)
+        F = Core.Typeof(f)
         source = methodinstance(F, tt)
         config = compiler_config(dev; name, kwargs...)::MetalCompilerConfig
         pipeline = GPUCompiler.cached_compilation(cache, source, config, compile, link)
@@ -197,7 +198,7 @@ function mtlfunction(f::F, tt::TT=Tuple{}; name=nothing, kwargs...) where {F,TT}
             kernel = HostKernel{F,tt}(f, pipeline)
             _kernel_instances[h] = kernel
         end
-        return kernel::HostKernel{F,tt}
+        return kernel
     end
 end
 
@@ -246,7 +247,7 @@ const _kernel_instances = Dict{UInt, Any}()
     ex
 end
 
-@inline function encode_argument!(kernel, arg)
+@inline function encode_argument!(@nospecialize(kernel), arg)
     argtyp = typeof(arg)
 
     # replace non-isbits arguments (they should be unused, or compilation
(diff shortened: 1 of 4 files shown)


maleadt:
Of course! I didn't have time to investigate the failure though.

src/compiler/execution.jl
```diff
@@ -179,11 +179,12 @@ The output of this function is automatically cached, i.e. you can simply call `m
 in a hot path without degrading performance. New code will be generated automatically when
 the function changes, or when different types or keyword arguments are provided.
 """
-function mtlfunction(f::F, tt::TT=Tuple{}; name=nothing, kwargs...) where {F,TT}
+function mtlfunction(@nospecialize(f), @nospecialize(tt)=Tuple{}; name=nothing, kwargs...)
```
christiangnrd: This despecialization breaks this inference test. But that seems to be on purpose so maybe remove?
https://github.com/JuliaGPU/Metal.jl/blob/1d2f000946cc60b3a9c2b6066734645c61b8c2dd/test/execution.jl#L45-L53

Only shows up when commenting out the `device_synchronize` test.

christiangnrd (commented):
I know this is still WIP, but hopefully this saves you some troubleshooting time.

`device_synchronize` seems to be broken due to the task-local storage call.

Edit: From a fresh session:
```julia-repl
julia> using Metal; length(Metal.global_queues)
1 # Should be 0
```

Edit 2: Adding `empty!(Metal.global_queues)` to `__init__()` seems to prevent the segfault, but that cannot be the proper solution right?

lib/mtl/storage_type.jl
```diff
@@ -46,9 +46,9 @@ See also [`Metal.SharedStorage`](@ref).
 struct PrivateStorage <: StorageMode end
 struct Memoryless     <: StorageMode end
 
-# Remove the ".MTL" when printing
-Base.show(io::IO, ::Type{<:PrivateStorage}) = print(io, "Metal.PrivateStorage")
-Base.show(io::IO, ::Type{<:SharedStorage}) = print(io, "Metal.SharedStorage")
+# Note: it'd be nicer to display these as `Metal.{Private,Shared}Storage` instead
+# of `Metal.MTL.{Private,Shared}Storage`, but defining `Base.show(::IO, ::Type{T})`
+# methods for that triggers thousands of invalidations against `Base.show(::IO, ::Type)`.
```
christiangnrd: What if we moved them into Metal proper instead? #779

PR #751, opened 2026-03-03
Error when attempting to launch kernels with an invalid grid size

christiangnrd:
Probably not caught because grid-stride loop kernels potentially still worked even with a truncated grid size due to the failure being silent, except when the total grid size was exactly `typemax(UInt32)`. Always fails shader validation.

Also shortens variable names like in #747


christiangnrd:
Why would the buildkite jobs not get picked up?

PR #768, opened 2026-04-17
Add pointer to integer conversion for alignment checking.

maleadt:


christiangnrd (approved):
I’m away from my computer so I can’t test locally but the tests seem good and [16 bytes alignment](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) seems like a reasonable assumption so lgtm!

PR #771, opened 2026-05-05
Run examples in isolated test processes

AshtonSBradley:
Examples may load packages that mutate global numerical-library state, such as AppleAccelerate. When examples are included directly in the package test runner, that state can leak into unrelated tests scheduled later on the same worker.

This keeps example coverage, but runs each example in a fresh Julia process with the example project active and TESTING=true. That makes the test suite independent of example load order and avoids example-only dependencies contaminating later package tests.

One example where this matters is #713, which adds FFT tests that should not depend on whether the flops comparison example loaded AppleAccelerate earlier on the same worker. The same isolation should help any future example that pulls in process-global state.

Type stability / allocation impact: this changes only how examples are run during tests. There is no public API or package runtime impact. Example tests now pay one Julia process startup per example in exchange for hermetic test state.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -87,19 +87,8 @@ for example in find_examples(joinpath(@__DIR__, "..", "examples"))
     name = splitext(basename(example))[1]
     dir = dirname(example)
     testsuite["examples/$name"] = quote
-        cd($dir) do
-            project = Base.active_project()
-            Base.set_active_project($dir)
-            try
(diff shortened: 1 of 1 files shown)


christiangnrd:
Thanks for your submission! I chose to approach this functionally similarly but targeting the problematic test set itself and taking advantage of ParallelTestRunner.jl functionality.

PR #780, opened 2026-05-24
Support for Metal 4 inline tensors + FlashAttention example

maleadt:
Somebody asked if Metal.jl supports the necessary primitives to implement flash attention, so I had Claude take a stab at it to see what's involved. Mostly vibe-coded, needs to be cleaned up and tested obviously.

---

This PR introduces a flash attention example, implemented in three ways:
- using basic MPS ops (matmul + softmax via broadcasts)
- using MPSGraph's fused `scaledDotProductAttention` op
- using a custom kernel built on `simdgroup_matrix intrinsics`

That needed a typed Julia wrapper around the existing simdgroup_matrix intrinsics (`MtlSimdgroupMatrix{T,8,8}`), plus a binding for `MPSGraph.scaledDotProductAttentionWithQueryTensor` (macOS 14+).

A fourth way would be to use the Metal 4 tensor ops (cooperative tensors + `tensor_ops::matmul2d`, with the softmax epilogue fused into the matmul via postfix fusion), but those require macOS 26, which I don't have yet. So to be continued...


maleadt:
I think this is getting close to ready. Not a perfect implementation yet, but a good starting point. Also only supports the inline tensors, passing tensors requires an ABI change that's much more involved.

lib/mpsgraphs/operations.jl
```diff
@@ -74,3 +74,28 @@ Dumps the `graph`.
...
+                                                  mask::MPSGraphTensor, scale::Real,
+                                                  name = "sdpa")
+    obj = @objc [graph::id{MPSGraph} scaledDotProductAttentionWithQueryTensor:Q::id{MPSGraphTensor}
+                                                            keyTensor:K::id{MPSGraphTensor}
+                                                          valueTensor:V::id{MPSGraphTensor}
+                                                           maskTensor:mask::id{MPSGraphTensor}
+                                                                scale:scale::Cfloat
+                                                                 name:name::id{NSString}]::id{MPSGraphTensor}
+    MPSGraphTensor(obj)
+end
```
christiangnrd: Minor nit. Can you move these above `dump_graph`? It's different from the rest in that it's not documented and for debugging only.

test/mpsgraphs/sdpa.jl
```diff
@@ -0,0 +1,85 @@
+if MPS.is_supported(device()) && Metal.macos_version() >= v"14"
```
christiangnrd: Could you add a warning that the tests are being skipped like we do in the profiling tests or filter it out in runtests? 
Also I recently removed the `MPS.is_supported` checks since we only support platforms that support it.

test/device/intrinsics/tensor.jl
```diff
@@ -0,0 +1,106 @@
+if Metal.macos_version() >= v"26"
```
christiangnrd: Same as my comment in spda

christiangnrd:
Only minor nits. The code itself seems fine afaict. I'm ambivalent towards the length of the comments since it feels like a lot, but for the most part they seem useful for getting a better low-level understanding.

src/tensor.jl
```diff
@@ -0,0 +1,90 @@
...
+
+Compute `C = A * B` (Julia matrix-product semantics) using
+`tensor_ops::matmul2d` with a tile decomposition. `A` is `(m, k)`, `B` is
+`(k, n)`, `C` is `(m, n)`, all column-major `MtlMatrix`. `C` is zeroed
+before the matmul (the kernel accumulates).
+
+`tile_m`, `tile_n`, `tile_k` set the per-threadgroup tile shape. The
+matrix dimensions must be evenly divisible by their respective tiles
+(`m % tile_m == 0`, `n % tile_n == 0`, `k % tile_k == 0`). Each
+threadgroup uses 4 SIMD-groups (128 threads on the M1/M2 hardware) and
```
christiangnrd: This seems like weird information. Is it not also 128 threads on newer hardware?

maleadt:
> 'm ambivalent towards the length of the comments since it feels like a lot

Yeah, I still want to do a pass condensing things and keeping only the useful ones, but it seemed valuable to keep them for an initial review.

That said, those breadcrumbs can be useful to help future humans and LLMs alike pick up on the design, even if they're sometimes a little long.

PR #788, opened 2026-05-28
Add nonblocking preference to versioninfo and LocalPreferences.toml

christiangnrd:
LocalPreferences string copied from CUDA, not sure if 100% correct for Metal.


christiangnrd:
This is all wrong

PR #798, opened 2026-06-02
Require 1.13 tests to pass

christiangnrd:
We missed a 1.13 regression introduced in #418 because it's allowed to fail.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -31,6 +31,7 @@ steps:
               - "1.10"
               - "1.11"
               - "1.12"
+              - "1.13"
 
   - group: ":soon: Pre-releases"
     steps:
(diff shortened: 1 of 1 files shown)


maleadt:
Should be possible after https://github.com/JuliaGPU/Metal.jl/pull/804

maleadt:
FWIW, the new downgrader supports LLVM 21 and 22, so we should be set for a while.

christiangnrd:
> FWIW, the new downgrader supports LLVM 21 and 22, so we should be set for a while.

Oh nice! I assumed you had just fixed the `va_arg` issue and didn't check.

christiangnrd:
Nightly error is https://github.com/JuliaLang/julia/pull/61762#issuecomment-4623506621

PR #808, opened 2026-06-05
Require macOS 14, and emit the newest supported AIR/MSL/metallib

maleadt:
Following Apple's policy of supporting three major versions, I had already bumped the README in https://github.com/JuliaGPU/Metal.jl/commit/eef641fd9a4806ef9d0f423fb4cc77c0beba6bbd. This PR now also updates the code, raising the minimum supported macOS version from 13 to 14. Code generation switches from conservative, pinned baselines (AIR 2.5, metallib v1.2.6) to the newest versions the host supports, which is what the offline `metal` compiler does: the deployment target determines the emitted AIR, MSL and metallib versions. Since we only ever JIT for the host device, there is no reason to target anything older.

Concretely, kernels now compile to AIR 2.6 / Metal 3.1 / metallib v1.2.7 on macOS 14, up to AIR 2.8 / Metal 4.0 / metallib v1.2.9 on macOS 26. That gives us access to newer language and IR features, and matches what Apple's toolchain produces; I verified the emitted versions against the offline compiler for every supported target.

Notable changes:

- Full read/write support for metallib v1.2.9 containers (Xcode 26), including the dynamic header. Libraries produced by any toolchain from macOS 11 through 26 now round-trip byte-exactly, which flushed out a few longstanding parsing quirks along the way. The container format is documented in the hacking FAQ.
- Generated libraries now get a content-derived UUID (the official toolchain stamps a random one), so compilation output is deterministic.
- The simdgroup matrix intrinsics moved to their generalized AIR 2.8 encoding. When targeting older macOS, the compiler rewrites them to the legacy encoding, so per-version targeting (`@metal macos=v"14" ...`) keeps working; codegen tests cover both encodings.
- Julia binaries built against a pre-Tahoe SDK see macOS 26 as 16; we now normalize such compatibility versions to the marketing version. Targeting an unsupported version errors instead of silently emitting something the OS may not load.


src/version.jl
```diff
@@ -1,13 +1,23 @@
...
+# Julia binaries built against a pre-Tahoe SDK observe compatibility versions (macOS 26
+# reports as 16); normalize to the marketing version, like the offline compiler
+# normalizes the deployment target.
+function normalize_macos(version::VersionNumber)
+    if v"16" <= version < v"26"
+        version = VersionNumber(version.major + 10, version.minor, version.patch)
+    end
+    return version
+end
+
```
christiangnrd, line 13: The ObjectiveC version already does this. https://github.com/JuliaInterop/ObjectiveC.jl/blob/7d0a7e18bd4e774cc16e06b501a8704a099b3a08/src/version.jl#L47-L55

Claude may have missed that because the docstring doesn't advertise it and I used `v"16"` instead of `v"26"` in the `x_support` functions below

maleadt: It's also used to normalize user-provided `macos` version strings.

maleadt:
Can't reproduce the perf differences locally. Probably noise from running on a different machine.

PR #815, opened 2026-06-08
Add a native GEMM

maleadt:
Adds Metal.jl's own matrix-multiply kernels so `A*B` / `mul!` no longer depend solely on
the vendor libraries (MPS / MPSGraph), and wires in a Metal 4 `tensor_ops::matmul2d` fast
path on capable hardware. Also makes the Metal 4 inline-tensor API idiomatic Julia.

## Matmul algorithms

Selected via `Metal.matmul_alg` (a `ScopedValue`):

- `:simd` — `simdgroup_matrix` 8x8 kernel, Float16/Float32 with F32 accumulate. Portable to
  every supported device. Handles any transpose / shape / offset / α,β.
- `:scalar` — per-element shared-memory tiled kernel. Any eltype (complex, bfloat16,
  integers), any transpose / offset. The correctness fallback.
- `:tensor` — Metal 4 `tensor_ops::matmul2d` (macOS 26+, Metal4-capable devices). Plain
  `C = A*B`, float eltypes, tile-divisible dims.
- `:native` — best of the three, auto-picked per device + operands (tensor → simd → scalar).
- `:auto` (default) — vendor first (MPSGraph → MPS where supported), else `:native`.
- `:MPS`, `:MPSGraph`, `:GPUArrays` — unchanged.

Each kernel has a matching `supports_<kind>_matmul` predicate (mirroring the existing
`supports_mps_matmul`), and `tensor_matmul_capable()` reports device support.

## Idiomatic inline-tensor API

The #780 `MtlInlineTensor` surface mirrored Apple's MSL conventions (row-major framing,
zero-based slice origins, operand/output transpose), forcing callers to reverse dimensions,
count from zero, and swap operands. The AIR primitive takes explicit extents/strides, so the
convention was ours to pick. Now column-major and 1-based:

- `MtlInlineTensor(A)` views all of `A` (its `size`); extents/strides are column-major.
- `view` takes 1-based origins.
- `matmul2d_descriptor(M, N, K)` describes a Julia `C[M,N] = A[M,K]*B[K,N]`;
  `transpose_left`/`transpose_right` transpose `A`/`B`.
- `op(A, B, C)` computes `C = A*B`. The swap onto Apple's convention happens inside the op
  at `@generated` time (constant-folded, zero runtime cost).

## Performance (GFLOP/s, macOS 26)

Apple M1 (no dedicated matmul units) vs M3 Pro (Apple9):

| eltype  | M×N×K | device |  MPS | MPSGraph | :simd | :tensor |
|---------|-------|--------|-----:|---------:|------:|--------:|
| Float32 | 512³  | M1     |  635 |      571 |   385 |     568 |
|         |       | M3 Pro |  570 |      553 |   590 |    1139 |
| Float32 | 2048³ | M1     | 1841 |     1727 |   575 |    1093 |
|         |       | M3 Pro | 4012 |     4006 |   731 |    3028 |
| Float32 | 4096³ | M1     | 1807 |     1763 |   549 |    1105 |
|         |       | M3 Pro | 4166 |     4151 |   708 |    3106 |
| Float16 | 4096³ | M1     | 1940 |     2280 |   564 |    1773 |
|         |       | M3 Pro | 4564 |     4565 |   743 |    3473 |

(`:GPUArrays`, the naive baseline: ~50 GFLOP/s on M1, ~100 on M3 Pro.)

Takeaways:

- The tensor path is the standout native kernel: ~2× the simdgroup kernel on M1, ~4–5× on
  M3 Pro.
- On M3 Pro it **beats the vendor libraries on small matrices** (512³ Float32: ~1140 vs
  ~560, about 2×), where MPS/MPSGraph launch overhead dominates.
- At large sizes it sustains ~75% of MPS/MPSGraph on M3 Pro (up from ~60% on M1). The vendor
  libraries stay ahead from 1024³ on up.
- The `:simd` kernel is tuned for M1 and does not scale on M3 Pro (flat ~0.7–0.9 TFLOP/s);
  on Metal 4 hardware the tensor path supersedes it.

This suggests `:auto` could route small float matmuls through `:tensor` instead of the
vendor libraries; the crossover is size- and eltype-dependent.


maleadt:
Benchmark script, in case other people want to run this:

```julia
using Metal
using LinearAlgebra: mul!
using Printf
using Base.ScopedValues: with

const ALGS  = (:MPS, :MPSGraph, :GPUArrays, :simd, :tensor)
const SIZES = (512, 1024, 2048, 4096)          # square M=N=K
const TYPES = (Float32, Float16)

# run C = A*B once under the given algorithm (errors if the algorithm can't handle it)
run_alg(alg, C, A, B) = with(() -> mul!(C, A, B), Metal.matmul_alg => alg)

# GFLOP/s for C = A*B under `alg`, or NaN if the algorithm is unsupported here. The timing
# is budgeted: fast algorithms get many iterations, slow ones (GPUArrays) only a few.
function gflops(alg, C, A, B; budget = 0.3, maxiters = 60, warmup = 2)
    try
        for _ in 1:warmup; run_alg(alg, C, A, B); end
        Metal.synchronize()
    catch
        return NaN
    end
    best = Inf; total = 0.0; n = 0
    while n < maxiters && total < budget
        t = @elapsed (run_alg(alg, C, A, B); Metal.synchronize())
        best = min(best, t); total += t; n += 1
    end
    M = size(C, 1); N = size(C, 2); K = size(A, 2)
    return 2.0 * M * N * K / best / 1e9
end

# sanity-check each algorithm against a CPU reference at a small size
function verify(T)
    M = N = K = 256
    A = MtlArray(rand(T, M, K)); B = MtlArray(rand(T, K, N))
    ref = Array(A) * Array(B)
    rtol = T === Float32 ? 1e-3 : 1.0f-1
    print("  verify $T:")
    for alg in ALGS
        C = MtlArray(zeros(T, M, N))
        status = try
            run_alg(alg, C, A, B); Metal.synchronize()
            isapprox(Array(C), ref; rtol) ? "ok" : "BAD"
        catch
            "n/a"
        end
        print(" ", alg, "=", status)
    end
    println()
end

function row(T, M, N, K)
    A = MtlArray(rand(T, M, K)); B = MtlArray(rand(T, K, N))
    @printf("%-7s %5d×%5d×%5d |", string(T), M, N, K)
    for alg in ALGS
        C = MtlArray(zeros(T, M, N))
        g = gflops(alg, C, A, B)
        isnan(g) ? @printf("  %9s", "-") : @printf("  %9.0f", g)
    end
    println()
end

function main()
    println("device: ", Metal.device())
    println("macOS:  ", Metal.macos_version(), "   Metal language: ", Metal.metal_support())
    println()
    for T in TYPES
        verify(T)
    end
    println()
    @printf("%-7s %19s |", "eltype", "M×N×K")
    for a in ALGS
        @printf("  %9s", a)
    end
    println("   (GFLOP/s)")
    println("-"^(19 + 21 + length(ALGS) * 11))
    for T in TYPES, s in SIZES
        row(T, s, s, s)
    end
end

main()
```

christiangnrd:
Some "fun" future work could be to implement/encode support for all type combinations listed in table 7.3 (MatMul2D data type supported) of the [metal shading language doc](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf)

PR #820, opened 2026-06-09
Default MtlArray storage to SharedStorage

KaanKesginLW:
## Summary

Change the default `MtlArray` storage mode from `PrivateStorage` to `SharedStorage`. On unified-memory (Apple Silicon) GPUs — essentially every supported Mac — `SharedStorage` is zero-copy between CPU and GPU and matches Apple's guidance. Discrete GPUs keep working (and get a one-time notice steering them to `PrivateStorage`).

```julia
using Metal
a = MtlArray(rand(Float32, 1024, 1024))   # SharedStorage by default now
w = unsafe_wrap(Array, a)                  # zero-copy CPU view — no Array() copy
```

This supersedes #717 with a clean, minimal diff (**+31 / −10** across 4 files, vs #717's +655 / −606).

## Motivation

- **Apple's guidance.** For unified memory, the [Metal Best Practices Guide](https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/MTLBestPracticesGuide/ResourceOptions.html) states *"the Shared mode is usually the correct choice."* The previous `PrivateStorage` default came from discrete-GPU guidance.
- **All supported Macs are Apple Silicon** (unified memory), so this is the right default for ~every user.
- **Zero-copy CPU access.** `SharedStorage` enables `unsafe_wrap(Array, x)`, avoiding the allocate-and-copy of `Array()`.

## Benchmarks

**No performance regression** (M2 Max) — `SharedStorage` ≈ `PrivateStorage`:

| size  | op        | Shared   | Private  |
|-------|-----------|---------:|---------:|
| 512²  | broadcast | 0.057 ms | 0.056 ms |
| 512²  | matmul    | 0.286 ms | 0.291 ms |
| 1024² | broadcast | 0.135 ms | 0.132 ms |
| 1024² | matmul    | 0.664 ms | 0.788 ms |

Consistent with the original 67-benchmark sweep: 78% ties, identical for `copyto!` / `fill!` / MPS-matmul, and all ties at ≥ 512 MB.

**Faster CPU access** via zero-copy `unsafe_wrap(Array, x)` instead of `Array()`
(M2 Max, best of 5; both paths `sum()` to force a full read):

| size   | `Array()` + use | `unsafe_wrap` + use | speedup |
|--------|----------------:|--------------------:|--------:|
| 512 MB | 17.1 ms  | 9.0 ms  | 1.9× |
| 1 GB   | 37.5 ms  | 18.2 ms | 2.1× |
| 2 GB   | 249.7 ms | 36.4 ms | 6.9× |
| 4 GB   | 539.2 ms | 72.5 ms | 7.4× |

(The jump at ≥ 2 GB is `Array()`'s GPU→CPU blit crossing into chunked copies;
`unsafe_wrap` stays flat because it never copies.)

## Implementation

- The `default_storage` preference now defaults to `"shared"`. Every constructor (`mtl`, `zeros`, `fill`, `similar`, …) already resolves through `DefaultStorageMode`, so they follow automatically.
- The default stays a **compile-time `const`** — type-stable and with **no device access during precompilation**. (It is *not* a `MTLDevice(…).hasUnifiedMemory` call baked into a precompiled constant.)
- On the rare **non-unified-memory (discrete) GPU**, `__init__` emits a one-time notice pointing the user to `set_preferences!(Metal, "default_storage" => "private")`. `SharedStorage` still works there; this just flags the faster option.
- Also fixes a latent bug in the old error message (`$default_storage` → `$str`).

## Notes

Supersedes #717, closed as *"this will probably eventually happen but it'll be easier to do it in a separate PR than to rebase and clean up this one."* This is that clean PR: no formatting churn, no test rewrite, just the default plus the discrete-GPU notice, docs, and a focused test.

Related: #717

### DIFF
--- a/docs/src/api/array.md
+++ b/docs/src/api/array.md
@@ -15,7 +15,7 @@ MtlVecOrMat
 
 ## Storage modes
 
-The Metal API has various storage modes that dictate how a resource can be accessed. `MtlArray`s are `Metal.PrivateStorage` by default, but they can also be `Metal.SharedStorage` or `Metal.ManagedStorage`. For more information on storage modes, see the official [Metal documentation](https://developer.apple.com/documentation/metal/resource_fundamentals/setting_resource_storage_modes).
+The Metal API has various storage modes that dictate how a resource can be accessed. `MtlArray`s are `Metal.SharedStorage` by default — zero-copy on unified-memory (Apple Silicon) GPUs — but they can also be `Metal.PrivateStorage` or `Metal.ManagedStorage`. On a discrete GPU, set the `default_storage` preference to `"private"` for best performance. For more information on storage modes, see the official [Metal documentation](https://developer.apple.com/documentation/metal/resource_fundamentals/setting_resource_storage_modes).
 
 ```@docs
 Metal.PrivateStorage
--- a/src/array.jl
+++ b/src/array.jl
@@ -39,7 +39,7 @@ end
 
 `N`-dimensional Metal array with storage mode `S` and elements of type `T`.
 
-`S` can be `Metal.PrivateStorage` (default), `Metal.SharedStorage`.
+`S` can be `Metal.SharedStorage` (default) or `Metal.PrivateStorage`.
 
 See the Array Programming section of the Metal.jl docs for more details.
 """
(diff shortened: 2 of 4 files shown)


christiangnrd:
#822

PR #821, opened 2026-06-09
Add support for BFloat16

### LINKED ISSUE #298: Add Support for BFloat16
With Julia 1.11 coming up, we will have native support for BFloat16 https://github.com/JuliaLang/julia/issues/41075.

Metal also supports BFloat16 onwards from [Apple6 (M1) GPU architecture](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf).

### LINKED ISSUE #817: [Julia 1.13] BFloat16 compilation failures
Prior to LLVM 19, on Apple Silicon, BFloat16 is treated as an i16. Julia 1.13 uses LLVM 20, which uses the LLVM bfloat type. This is causing compilation failures on native kernels using BFloat16.

Seems to be the reason for test failures on 1.13 for https://github.com/JuliaGPU/Metal.jl/pull/812 and https://github.com/JuliaGPU/Metal.jl/pull/446.

maleadt:
Julia 1.13+ only, since that's the only release where we generate `bfloat` on AArch64 (even though this is only a property of the CPU back-end...).

Fixes https://github.com/JuliaGPU/Metal.jl/issues/298
Fixes https://github.com/JuliaGPU/Metal.jl/issues/817


maleadt:
Turns out that software-emulated BFloat16 already worked. I wonder if we should warn users about this, but it's tricky, since our current bad datatype detection operates at the LLVM level (and software-emulated BFLoat16 is simply `i16` there).

PR #822, opened 2026-06-09
Default to SharedStorage

christiangnrd:


src/array.jl
```diff
@@ -39,7 +39,8 @@ end
 
 `N`-dimensional Metal array with storage mode `S` and elements of type `T`.
 
-`S` can be `Metal.PrivateStorage` (default), `Metal.SharedStorage`.
+`S` can be `Metal.SharedStorage` (default), `Metal.PrivateStorage`. The default
+storage mode can be changed by setting `default_storage` in LocalPreferences.
```
christiangnrd:
```suggestion
storage mode can be changed by setting `default_storage` in LocalPreferences.toml.
```

christiangnrd:
Could the github actions [failure](https://github.com/JuliaGPU/Metal.jl/actions/runs/27759897413/job/82131223290) be related to the compilation failures?

maleadt:
I'm not sure how. There's a known issue with ObjC errors leaking outside of their retain/release scope and crashing during error reporting, but that's read-only and shouldn't cause a crash in LLVM.

maleadt:
I guess the remaining question is semantics. Do we want to allow scalar iteration on shard memory so that it can be used with CPU code? It'll never be as fast as Array, but if we port the `dirty` memory flag tracking from CUDA.jl we can get it down to a couple of ns (as opposed to ~1ns for an Array getindex or so). If we want to add back scalar iteration checking we add another couple of ns for the TLS check on every access.

christiangnrd:
i.e. `unsafe_wrap(Array, ...` wouldn't be needed?

maleadt:
Right, that's the current design of CUDA.jl:

```julia-repl
Precompiling CUDA finished.
  12 dependencies successfully precompiled in 43 seconds. 90 already precompiled.

julia> CUDA.allowscalar(false)

julia> a = cu([1])
1-element CuArray{Int64, 1, CUDACore.DeviceMemory}:
 1

julia> a[]
ERROR: Scalar indexing is disallowed.

julia> b = cu([1]; unified=true)
1-element CuArray{Int64, 1, CUDACore.UnifiedMemory}:
 1

julia> b[]
1
```

I'm not entirely convinced this is the best option though. It makes sense, but people often use `allowscalar` for detecting GPU code. We could tell them they have to use private memory for that, but it still makes `allowscalar(false)` a lie.

christiangnrd:
This is also how shared storage currently works

christiangnrd:
But without the `dirty` flag so there might be some latent race conditions with shared MtlArrays

maleadt:
> This is also how shared storage currently works

Right, but it never was the default. It would break users doing `allowscalar(false)` to detect GPU functionality execution on the CPU.

And we definitely need the `dirty` flag to improve performance here. But that can be follow-up work.

christiangnrd:
What of we add a `@warn` when they call allowscalar the first time if default storage mode is shared?

maleadt:
I guess that could work, but the interface is not extensible like that right now.

EDIT: I guess we could implement `Metal.allowscalar`, have it warn, and then call `GPUArrays.allowscalar`.

christiangnrd:
New relevant case: unsupported sorts (`by != identity`) will silently run on the CPU under SharedStorage.

PR #824, opened 2026-06-12
Update CI

christiangnrd:
Add explicit "m1" tag to benchmarks to avoid issues if we ever get runners on newer GPUs


maleadt:
If we're using the JuliaGPU cluster, the secure pipeline and everything to do with cryptic can go.

maleadt:
CI is going to take ages though on my single M2...

christiangnrd:
Are the old M1 runners only in the julia queue until the juliaecosystem queue is recreated?

Also this runner reports being an M1

maleadt:
Yes I was mistaken; it's an M1. Once I have a better picture of the state of Base CI, I'll claw back some agents for JuliaGPU. And maybe buy a couple when the M5 mac mini releases.

> Are the old M1 runners only in the julia queue until the juliaecosystem queue is recreated?

`juliaecosystem` is unlikely to come back

christiangnrd:
~Not sure why the Benchmarks aren't waiting for every other test to be done before running.~
Edit: they depend on successful completion of the main tests

christiangnrd:
The only related issue remaining is Documenter failure which Iprobably requires more permissions to address.

Other failures:
- On 1.10 and 1.11, the tensor matmul errors are #825 
- The output/locale issue seems to have been magically fixed on 1.13 but not 1.12. Assuming the locale has been changed would clearing the caches potentially fix this on 1.12?
- API Validation failures are #827
- The `attention_tensor` function in the flash_attention example fails under shader validation for using too much threadgroup memory. I've opened #828 to guard against this, but the example still needs to be fixed (or disabled under shader validation?)

PR #835, opened 2026-06-16
Fix broadcasting of large arrays under shader validation

christiangnrd:


christiangnrd:
This isn't tested on CI and I just want the benchmarks

PR #837, opened 2026-06-17
Keep tensor op descriptors out of constant AS

### LINKED ISSUE #834: Tensor gemm shader validation failures
The tensor gemm failures are real. The decomposition ones false positives caused by the gpu refusing new commands after the tensor failures.

See: https://buildkite.com/julialang/metal-dot-jl/builds/2910#019ed081-e478-4b15-8c2a-cdaa17335b91

maleadt:
Metal Shader Validation faults when __tensorops_impl_matmul2d_op_run receives a matmul descriptor initialized from a global that GPUCompiler cloned into AIR constant address space 2. GPUCompiler now provides a Metal hook for selecting the address space of each constant global, so override it for globals that feed tensor-op descriptor copies and leave only those constants in AS0 while preserving the default AS2 lowering for ordinary constants.

The inline tensor constructors also mark their explicitly packed-stride descriptors as non-contiguous, matching the initialization path used by the tensor-op runtime.

Fixes https://github.com/JuliaGPU/Metal.jl/issues/834


maleadt:
Looks like https://github.com/JuliaGPU/GPUCompiler.jl/pull/848 caused issues here.

christiangnrd:
Nightly failure is https://github.com/JuliaGPU/GPUArrays.jl/pull/735

maleadt:
We triggered a different validation failure first, I think:
```
Error in testset examples/flashattention:
Error During Test at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-macmini-aarch64-2.1/depots/5cd495a2-4a16-4674-ae02-c839447744bb/packages/ParallelTestRunner/kZFbj/src/ParallelTestRunner.jl:400
  Got exception outside of a @test
  LoadError: ArgumentError: Total used threadgroupMemoryLength(49152) must be <= 32768 bytes.
```

christiangnrd:
The main :tensor tests weren't run but the flopscomp example (which ran first) tests the tensor (via :native) path as of my changes yesterday so this should be good. The flashattention failure I mention in https://github.com/JuliaGPU/Metal.jl/pull/824#issuecomment-4696647001 and I'm pretty sure it's the last validation error for CI to be green again

PR #841, opened 2026-06-17
Dump artifacts when LLVM generation failed.

maleadt:
For https://github.com/JuliaGPU/GPUCompiler.jl/issues/851

Let's retry builds here until we catch the bug


maleadt:
I can't get this to reproduce, so let's merge instead.

PR #844, opened 2026-06-19
Work around REPL invalidating precompilation

christiangnrd:
Seems to be 1.14-only


maleadt:
Hmm, I'd rather we investigate this first. The whole premise of https://github.com/JuliaLang/julia/pull/61714 was to _avoid_ needing this.

maleadt:
I've added this to GPUCompiler

PR #847, opened 2026-06-20
Add a batched command queue to amortize launch overhead

maleadt:
To amortize the cost of creating and submitting a command buffer on every launch, make `global_queue` return a batched command queue that allows submitting tasks to an open command buffer, automatically flushed and submitted on synchronization or certain other events.


maleadt:
Yay, some nice speed-ups in there.

PR #855, opened 2026-06-22
Use MPS for accumulate/reduce/sort

maleadt:
Sort didn't work before, and accumulate/reduce are a bit faster in some cases, and never slower. At least not right now, probably will have to re-evaluate once we switch to AK.jl.


maleadt:
Some impressive speed-ups!

PR #863, opened 2026-06-23
More types for MPSGraph reduction/scan/sort support

christiangnrd:
Funny enough I tried to add `&` and `|` reductions but Base is broken for those until https://github.com/JuliaLang/julia/pull/58418

### DIFF
--- a/lib/mpsgraphs/scan.jl
+++ b/lib/mpsgraphs/scan.jl
@@ -1,4 +1,4 @@
-const MPSGRAPH_VALID_SCAN_TYPES = Union{Float16, Float32}
+const MPSGRAPH_VALID_SCAN_TYPES = filter(T -> T <: Real && T != Bool, (MPS.jl_mps_to_typ |> values |> collect))
 
 function cumulativeSumWithTensor(graph::MPSGraph, tensor::MPSGraphTensor,
                                  axis::Integer, exclusive::Bool,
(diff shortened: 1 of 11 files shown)


lib/mpsgraphs/scan.jl
```diff
@@ -40,21 +40,27 @@ function cumulativeMinimumWithTensor(graph::MPSGraph, tensor::MPSGraphTensor,
...
-scan_operation(::typeof(*)) = :product
-scan_operation(::typeof(Base.mul_prod)) = :product
-scan_operation(op) =
-    throw(ArgumentError("MPSGraph scan supports + and *"))
+scan_operation(::DataType, ::typeof(+)) = :sum
+scan_operation(::DataType, ::typeof(Base.add_sum)) = :sum
+scan_operation(::DataType, ::typeof(*)) = :product
+scan_operation(::DataType, ::typeof(Base.mul_prod)) = :product
+scan_operation(::Type{<:Integer}, ::typeof(min)) = :minimum
+scan_operation(::Type{<:Integer}, ::typeof(max)) = :maximum
```
maleadt, line 48: Yeah I had removed these because of incompatible NaN propagation, but you're right that it should work for integers.

christiangnrd:
Edit: I've decided `Bool` for scans and reductions is future work for someone who thinks they would benefit from it
~A few quirks I'm probably overthinking:~
~- I don't include `Bool` for `scan` when it's technically supported and does something for min/max~
~- bitwise and and or reductions are broken in Base so I didn't add them butmaybe I should?~
~- Properly encoding support for types/operations will become a mess of lists/unions. Is it worth it?~

christiangnrd:
3x regression for the `array/accumulate/Int64/dims=1L` benchmark. Other Int64 benchmarks seem to be the same or improved.

Issue #868, opened 2026-06-30
macOS 27: tensor-ops gemm test failures


christiangnrd:
```Julia-repl
     Testing Running tests...
┌ Warning: Metal.jl has not been tested on macOS 27 or later, you may run into issues.
└ @ Metal ~/.julia/dev/Metal/src/initialization.jl:67
┌ Info: System information:
│ macOS 27.0.0, Darwin 27.0.0
│
│ Toolchain:
│ - Julia: 1.12.7-DEV.5
│ - LLVM: 18.1.7
│ - Metal: 4.1 (MSL), 2.9 (AIR), 1.2.9 (metallib)
│
│ Julia packages:
│ - Metal.jl: 1.10.0
│ - GPUArrays: 11.5.8
│ - GPUCompiler: 1.22.5
│ - KernelAbstractions: 0.9.42
│ - ObjectiveC: 6.0.0
│ - LLVM: 9.10.0
│ - LLVMDowngrader_jll: 0.8.1+0
│
│ 1 device:
└ - Apple M2 Max (30 GPU cores, 64.000 KiB allocated; Apple8, Metal4 family)
Running 1 tests using 1 parallel jobs. If this is too many concurrent jobs, specify the `--jobs=N` argument to the tests, or set the `JULIA_CPU_THREADS` environment variable.
Available memory: 22.516 GiB
               │   Test   │   Init   │ Compile │ ──────────────── CPU ──────────────── │
Test  (Worker) │ time (s) │ time (s) │   (%)   │ GC (s) │ GC % │ Alloc (MB) │ RSS (MB) │
linalg     (1) │        started at 2026-06-30T09:23:34.425
linalg     (1) │    82.83 │     6.79 │    failed at 2026-06-30T09:24:58.033

Output generated during execution of 'linalg':
┌ ┌ Warning: Metal.jl has not been tested on macOS 27 or later, you may run into issues.
└ └ @ Metal ~/.julia/dev/Metal/src/initialization.jl:67

Test Summary:                                       | Pass  Fail  Total     Time
  Overall                                           |  360     4    364  1m30.9s
    linalg                                          |  360     4    364  1m22.8s
      matmul algorithm selection                    |   46           46    17.7s
      test matrix vector multiplication of views    |    2            2     0.1s
      matrix multiplication of range views          |    7            7     2.0s
      native GEMM                                   |  212     4    216    41.2s
        Float32 NN α=1.0 β=0.0                      |    2            2     0.5s
[...]

PR #869, opened 2026-06-30
Disable automatic selection of tensor gemm on macos 27

christiangnrd:
Mitigates #868 until a proper fix is implemented

Not in `supports_tensor_matmul` because I just want to disable the default route


maleadt:
Hmm, I can't reproduce this on a virtual macOS 27... So LGTM I guess.

christiangnrd:
> Hmm, I can't reproduce this on a virtual macOS 27...

Neither can I

PR #870, opened 2026-06-30
Add ordered atomics

VEZY:
Note: I used Codex with GPT5.5 High to implement this feature. For context, my original issue with this problem was about building a correct TLAS on the GPU with RayCore.jl (which failed): https://github.com/JuliaGeometry/Raycore.jl/pull/19

## Summary

This adds explicit memory-order support to Metal’s low-level atomic intrinsics and enables KernelAbstractions atomic support for `MetalBackend`.

Metal atomic operations previously exposed only relaxed ordering through the Julia API. That is fine for counters and reductions, but it is not enough for synchronization patterns where ordinary device-memory writes are published through an atomic operation and then read by another workitem.

Changes:
- Add Metal `memory_order` enum values for relaxed, acquire, release, acq_rel, and seq_cst.
- Add explicit ordering arguments to low-level atomic load/store/exchange/fetch/CAS functions.
- Preserve relaxed defaults for existing direct Metal atomic call sites.
- Add compare-exchange success/failure ordering support.
- Use `Val`-specialized ordering paths so GPU code receives compile-time ordering constants.
- Enable `KernelAbstractions.supports_atomics(::MetalBackend)`.
- Add a regression test for a cross-workitem publish/consume pattern using device-memory writes plus an ordered atomic `fetch_add`.

## Why

KernelAbstractions and Atomix can express atomic operations with acquire/release or sequentially-consistent semantics. Metal needs to receive those requested orderings for synchronization-sensitive GPU algorithms to be correct.

The new regression test models a common bottom-up refit pattern: each workitem writes ordinary device memory, then increments an atomic counter; the second arriving workitem reads both children’s ordinary writes and publishes the parent. This requires the atomic operation to provide ordering, while still allowing relaxed atomics for cases that do not need synchronization.

## Validation

- Metal atomic tests passed.
- Metal KernelAbstractions integration test passed.
- Downstream RayCore TLAS/BLAS Metal repro passed with Atomix orderings wired through to Metal.
- ArchimedLight-derived static TLAS comparison passed: CPU-built/adapted and direct Metal-built TLAS match for nodes, instances, all BLAS buffers, descriptors, and root AABB.

### DIFF
--- a/src/MetalKernels.jl
+++ b/src/MetalKernels.jl
@@ -33,7 +33,7 @@ KA.synchronize(::MetalBackend) = synchronize()
 KA.functional(::MetalBackend) = Metal.functional()
 
 KA.supports_float64(::MetalBackend) = false
-KA.supports_atomics(::MetalBackend) = false
+KA.supports_atomics(::MetalBackend) = true
 KA.supports_unified(::MetalBackend) = true
 
 Adapt.adapt_storage(::MetalBackend, a::Array) = Adapt.adapt(MtlArray, a)
--- a/src/MetalKernels.jl
+++ b/src/MetalKernels.jl
@@ -33,7 +33,7 @@ KA.synchronize(::MetalBackend) = synchronize()
 KA.functional(::MetalBackend) = Metal.functional()
 
 KA.supports_float64(::MetalBackend) = false
-KA.supports_atomics(::MetalBackend) = true
+KA.supports_atomics(::MetalBackend) = metal_support() >= v"4.1"
 KA.supports_unified(::MetalBackend) = true
 
 Adapt.adapt_storage(::MetalBackend, a::Array) = Adapt.adapt(MtlArray, a)
--- a/src/device/intrinsics/atomics.jl
+++ b/src/device/intrinsics/atomics.jl
@@ -1,37 +1,5 @@
 # Atomic Functions
 
-@enum memory_order::Int32 begin
-    memory_order_relaxed = 0
-    memory_order_acquire = 2
-    memory_order_release = 3
-    memory_order_acq_rel = 4
-    memory_order_seq_cst = 5
-end
-
-# Ordered atomics are only available in newer Metal language versions. Keep this
-# check in device code so target constants can propagate and an enclosing
-# `metal_version()` guard can eliminate an otherwise unsupported operation.
-@inline function check_memory_order(::Val{order}) where {order}
-    if order === memory_order_relaxed
-        return
-    elseif order === memory_order_seq_cst
-        @static_assert(metal_version() >= sv"4.0",
-                       "Atomic memory_order_seq_cst requires Metal 4.0 or newer.")
-    elseif order === memory_order_acquire
-        @static_assert(metal_version() >= sv"4.1",
-                       "Atomic memory_order_acquire requires Metal 4.1 or newer.")
-    elseif order === memory_order_release
-        @static_assert(metal_version() >= sv"4.1",
-                       "Atomic memory_order_release requires Metal 4.1 or newer.")
-    elseif order === memory_order_acq_rel
-        @static_assert(metal_version() >= sv"4.1",
-                       "Atomic memory_order_acq_rel requires Metal 4.1 or newer.")
-    else
-        @static_assert(false, "Invalid atomic memory ordering.")
-    end
-    return
-end
-
 # XXX: the integers should come from some enum
 const atomic_memory_names = Dict(
     AS.Device      => ("global", Int32(2)),
@@ -51,6 +19,7 @@ const atomic_type_names = Dict(
 for typ in (:Int32, :UInt32), as in (AS.Device, AS.ThreadGroup)
     typnam = atomic_type_names[typ]
     memnam, memid = atomic_memory_names[as]
+    default_flags = as === AS.Device ? MemoryFlagDevice : MemoryFlagThreadGroup
 
     @eval begin
         @inline atomic_store_explicit(ptr::LLVMPtr{$typ,$as}, desired::$typ) =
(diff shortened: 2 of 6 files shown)


VEZY:
Of course I also kept the default memory order to relaxed.

VEZY:
I put `supports_atomics` to `true`:

```julia
KA.supports_atomics(::MetalBackend) = true
```

because I also opened a PR on Atomix here: https://github.com/JuliaConcurrent/Atomix.jl/pull/79

maleadt:
Even though these orderings seemed to work, both on CI and on my system, I prefer if we follow the Apple spec: 
image

To make it easier to write code against different MSL versions, I added https://github.com/JuliaGPU/GPUCompiler.jl/pull/880 and applied that here.

christiangnrd:
FYI there seems to be a typo in Apple's table. `memory_order_seq_cst` was added in Metal 3.2 for `atomic_thread_fence` only and that doesn't seem to have changed in Metal 4

src/device/intrinsics/atomics.jl
```diff
@@ -2,6 +2,34 @@
...
+end
+
+# Ordered atomics are only available in newer Metal language versions. Keep this
+# check in device code so target constants can propagate and an enclosing
+# `metal_version()` guard can eliminate an otherwise unsupported operation.
+@inline function check_memory_order(::Val{order}) where {order}
+    if order === memory_order_relaxed
+        return
+    elseif order === memory_order_seq_cst
+        @static_assert(metal_version() >= sv"4.0",
```
christiangnrd: Before Metal 4.1, `memory_order_seq_cst` was only supported with `atomic_thread_fence` which we haven't wrapped yet.
```suggestion
        @static_assert(metal_version() >= sv"4.1",
```

test/device/intrinsics/atomics.jl
```diff
@@ -228,6 +228,42 @@ n = 128 # NOTE: also hard-coded in MtlThreadGroupArray constructors
...
+        function ordered_fetch_kernel(a, ::Val{ORDER}) where {ORDER}
+            Metal.atomic_fetch_add_explicit(pointer(a, 1), Int32(1), ORDER)
+            return
+        end
+
+        # The enum value is in the kernel signature, so the public enum overload
+        # specializes it to a Val before lowering to AIR.
+        for (order, minimum, message) in (
+            (Metal.memory_order_relaxed, v"3.0", nothing),
+            (Metal.memory_order_seq_cst, v"4.0",
```
christiangnrd: See previous comment
```suggestion
            (Metal.memory_order_seq_cst, v"4.1",
```

christiangnrd (changes requested):
This needs a minor rewrite. `memory_order_relaxed` is still the only _supported_ order for all released versions of macOS, and with the unreleased Metal 4.1, atomics lower to a different form with an extra argument for the memory flag

maleadt:
Can you test on macOS 27? I only have a virtualized environment.

christiangnrd:
Tests seem to pass on macOS 27 except 1.10, but I can't help but feel like there's a much less messy way to do this. Considering the atomic operation changes really only apply to macOS 27, I'll split off the fence stuff into its own PR

maleadt:
Okay, different approach, targeting a single IR form and downgrading during compilation. That's similar to what we do for simdgroup intrinsics. WDYT?

christiangnrd:
> targeting a single IR form and downgrading during compilation

That was one reason for my hesitation about the previous method. AIR 2.9 always lowers to the extra-arg version. I assume the old method is accepted for compatibility but I'd rather do it right since we never know when they'll get around to adding it to the validator.

My other gripe was about the sheer amount of code duplication which this new approach also seems to improve upon. I still think we should merge #875 separately first but at least now the diff is manageable for me to review

src/device/intrinsics/atomics.jl
```diff
@@ -19,60 +15,172 @@ const atomic_type_names = Dict(
...
+            flags::Union{MemoryFlags,UInt32}=default_atomic_flags(order, Val($as)))
+            atomic_load_explicit(ptr, Val(order), Val(flags))
         end
 
-        function atomic_load_explicit(ptr::LLVMPtr{$typ,$as})
+        function atomic_load_explicit(ptr::LLVMPtr{$typ,$as}, ::Val{O},
+                                      ::Val{F}) where {O,F}
+            @static_assert(O === memory_order_relaxed || O === memory_order_acquire ||
+                           O === memory_order_seq_cst,
+                           "atomic_load_explicit only supports relaxed, acquire, or sequentially consistent ordering.")
```
christiangnrd: Not sure this is true

src/device/intrinsics/atomics.jl
```diff
@@ -19,60 +15,172 @@ const atomic_type_names = Dict(
...
+    order === memory_order_seq_cst
+
+# MSL 4.1 requires callers to spell out flags for ordered atomics. We instead default
+# to the memory region addressed by the pointer, while retaining mem_none for relaxed
+# operations to match the no-flags MSL overload.
+@inline atomic_order_and_flags_available(order, flags) =
+    (order === memory_order_relaxed && flags == MemoryFlagNone) ||
+    metal_version() >= sv"4.1"
+
+# MSL specification section 6.15: failure ordering is restricted to relaxed, acquire,
```
christiangnrd: Apple doesn't care about section name "ABI" compatibility. This is already outdated when looking at the Metal 4.1 version. I wouldn't specify a section number

src/device/intrinsics/atomics.jl
```diff
@@ -95,23 +203,47 @@ for (op, types) in atomic_fetch_and_modify, typ in types, as in (AS.Device, AS.T
...
+
+        function $f(ptr::LLVMPtr{$typ,$as}, desired::$typ, ::Val{O}, ::Val{F}) where {O,F}
+            @static_assert(valid_atomic_order(O), "Invalid atomic memory ordering.")
+            @static_assert(atomic_order_and_flags_available(O, F),
+                           "Ordered atomics and memory flags require Metal 4.1 or newer.")
             @typed_ccall($"air.atomic.$memnam.$op.$typnam", llvmcall, $typ,
-                         (LLVMPtr{$typ,$as}, $typ, Int32, Int32, Bool),
-                         ptr, desired, Val(memory_order_relaxed), Val($memid), Val(true))
+                         (LLVMPtr{$typ,$as}, $typ, Int32, Int32, Int32, Bool),
+                         ptr, desired, Val(O), Val($memid), Val(F), Val(false))
```
christiangnrd: I guess this isn't a new issue but Float32 `add` and `sub` for threadgroup memory aren't supported before metal 4.1

src/device/intrinsics/atomics.jl
maleadt: I was following C++14-memory-model UB rules here. But I'll relax it.

src/MetalKernels.jl
```diff
@@ -33,7 +33,7 @@ KA.synchronize(::MetalBackend) = synchronize()
 KA.functional(::MetalBackend) = Metal.functional()
 
 KA.supports_float64(::MetalBackend) = false
-KA.supports_atomics(::MetalBackend) = false
+KA.supports_atomics(::MetalBackend) = true
```
christiangnrd:
```suggestion
KA.supports_atomics(::MetalBackend) = metal_support() >= v"4.1"
```

test/kernelabstractions.jl
```diff
@@ -1,5 +1,8 @@
 import KernelAbstractions
 using Metal.MetalKernels
+using Test
+
+@test KernelAbstractions.supports_atomics(MetalBackend())
```
christiangnrd: This or just remove? It doesn't feel like a very useful test

```suggestion
@test KernelAbstractions.supports_atomics(MetalBackend()) == Metal.metal_support() >= v"4.1"
```

src/device/intrinsics/atomics.jl
```diff
@@ -1,113 +1,194 @@
...
-    :UInt64  => "i64",
-    :Float32 => "f32"
-)
 
+@inline function default_atomic_flags(order::memory_order, ::Val{A}) where {A}
+    order === memory_order_relaxed && return MemoryFlagNone
+    A === AS.Device ? MemoryFlagDevice : MemoryFlagThreadGroup
+end
+
+@inline valid_atomic_order(order) =
```
christiangnrd: Can't this be represented as cleanly without the need for a function with `order isa memory_order`?

christiangnrd:
Is https://github.com/JuliaGPU/Metal.jl/commit/5598ff257758057fa6e607f1be10487fbef71ac9 still here on purpose?

maleadt:
> Is [5598ff2](https://github.com/JuliaGPU/Metal.jl/commit/5598ff257758057fa6e607f1be10487fbef71ac9) still here on purpose?

No, that's something I reverted on main, and apparently snuck in here again.

christiangnrd:
I lost the more detailed comment I had typed up, but I think I confused the robot when merging #875 and force-pushing to update this PR and it's somehow back to the messy version that made me comment https://github.com/JuliaGPU/Metal.jl/pull/870#issuecomment-4994494416

This PR shouldn't need any code or test changes to atomic fences unless we decide we want the atomic check functions in synchronization. But I had removed `check_atomic_thread_fence_order` in #875 since it was only used once and didn't clutter up the function too much so the other ones could probably be moved to atomics.jl

maleadt:
Github is being confused here... I pushed a clean-up commit, but it doesn't appear. I'll re-open on a first-party branch.

maleadt:
https://github.com/JuliaGPU/Metal.jl/pull/889

VEZY:
Thanks @maleadt and @christiangnrd for pushing this forward ! And sorry for not having participated more in the discussion; that was way out of my capabilities.

PR #876, opened 2026-07-16
Avoid checked conversions in tensor GEMM kernels

maleadt:


maleadt:
FWIW, this shouldn't have been required. I'm not sure where the sudden failure came from, but it now also happens on FlashAttention. Maybe a bug/change in Julia or GPUCompiler? I'm taking a look.

christiangnrd:
I didn’t even notice a failure. What configuration/version was it happening with?

maleadt:
It's an order-dependent inference failure, observed on 1.12+. I also noticed it on #870 where the FlashAttention example started to fail.

PR #885, opened 2026-07-21
MacOS 27 support

christiangnrd:
Follow-up PR will remove macOS 14 support.


christiangnrd:
All tests pass locally on macOS 27 RC normally and under shader validation!

PR #889, opened 2026-07-23
Add ordered atomics and enable KernelAbstractions atomic support

maleadt:
Metal's low-level atomic intrinsics previously exposed only relaxed ordering. That is fine for counters and reductions, but not for synchronization patterns where ordinary device-memory writes are published through an atomic and then observed by another workitem.

Thread explicit memory orderings through the low-level atomic API and light up KernelAbstractions/Atomix atomics on the Metal backend:

- Add explicit `order` (and `flags`/`scope`) arguments to the low-level atomic store/load/exchange/fetch/compare-exchange intrinsics, using `Val`-specialized paths so orderings reach GPU code as compile-time constants.
- Keep the existing direct call sites on relaxed ordering, so code that only needs relaxed atomics keeps emitting the same `air.atomic.*` forms and does not change behaviour.
- Support independent success/failure orderings on compare-exchange.
- Gate the acquire/release/acq_rel/seq_cst orderings and the memory flags on Metal 4.1 with device-side `@static_assert`s (`check_atomic_memory_order` / `check_atomic_flags`), so unavailable operations fail to compile with a clear message and enclosing `metal_version()` guards can eliminate them.
- Enable `KA.supports_atomics(::MetalBackend)` when the device supports Metal 4.1, so Atomix-based kernels (e.g. KernelAbstractions' `@atomic`) work on Metal.

### DIFF
--- a/src/MetalKernels.jl
+++ b/src/MetalKernels.jl
@@ -33,7 +33,7 @@ KA.synchronize(::MetalBackend) = synchronize()
 KA.functional(::MetalBackend) = Metal.functional()
 
 KA.supports_float64(::MetalBackend) = false
-KA.supports_atomics(::MetalBackend) = false
+KA.supports_atomics(::MetalBackend) = metal_support() >= v"4.1"
 KA.supports_unified(::MetalBackend) = true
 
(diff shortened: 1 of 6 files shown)


christiangnrd (approved):
I think this is ready

PR #890, opened 2026-07-23
Recover lost atomic ordering changes

christiangnrd:
I think I was able to recover the lost commits and changes. The last commit is changes that happened between the force-push that lost these and the previous rebase to main so it's a mix of review changes and a bit of cleanup from me.


maleadt:
Awesome, thanks!

PR #906, opened 2026-07-30
Remove broken `copy(::MPS.RNG)`

christiangnrd:
This was broken from `MTLDevice` not having a `copy` available to it, but it doesn't copy the underlying random kernel state so it's kind of pointless.


christiangnrd:
Will merge. If this needs the be reverted, don't revert the test since that's to add coverage for that seeding entry point

PR #920, opened 2026-08-18
Handle transposed triangular operands in in-place left division.

maleadt:
`ldiv!` lacked the transposed-operand method its `rdiv!` sibling already has, so `transpose(UpperTriangular(::MtlMatrix)) \ x` fell back to the generic `ldiv!`, whose `istriu` direction check scalar-indexes the wrapper. Latent until 1.12.7, where LinearAlgebra.jl#1561 routed `\` through the in-place `ldiv!`; 1.13 avoids it via LinearAlgebra.jl#1265.


maleadt:
Oh wait we can just do this in GPUArrays; https://github.com/JuliaGPU/GPUArrays.jl/pull/763

PR #928, opened 2026-08-21
add override for `Bool` and intercept `throw(InexactError(...))`

simeonschaub:
Both this PR and https://github.com/JuliaGPU/GPUCompiler.jl/pull/904 fix the issue reported in https://github.com/JuliaGPU/GPUCompiler.jl/pull/899#issuecomment-5372635495 on their own, but both are separate issues that just surfaced together to cause these errors

Adds a device override for `Bool(::Real)` and `Bool(::Float16)` that uses `@gpu_throw` for the error path instead of throwing a regular `InexactError`. We already do this for `Core.throw_inexacterror`, but the `Bool` constructor doesn't call that function


christiangnrd (approved):
Thanks!

PR #930, opened 2026-08-24
Enable persistent Metal pipeline caching by default

maleadt:
GPUCompiler 2.3 produces deterministic Metal libraries across Julia sessions, making Metal binary archives reliable enough to enable by default. This allows kernels compiled in one session to reuse their native GPU code in later sessions, avoiding repeated Metal driver compilation.

Archives are stored independently per kernel and grouped by GPU and operating-system build. This keeps memory usage proportional to the kernels actually used, avoids repeatedly rewriting a growing shared archive, and allows individual entries to be evicted safely. Archive writes are immediate and atomic, stale caches are cleaned automatically, and any archive failure falls back to normal compilation.

The cache has a configurable disk-size limit and reports session hit/miss statistics through Metal.versioninfo(). Tests cover cross-process reuse, corrupt and stale archives, cache eviction, relocation-carrying kernels, precompilation, and shader validation across all supported Julia versions.


src/compiler/archive.jl
```diff
@@ -49,16 +53,33 @@ const archive_hits = Threads.Atomic{Int}(0)
...
+    Base.@lock device_archives_lock empty!(device_archives)
+    archive_hits[] = 0
+    archive_misses[] = 0
+    _os_build[] = nothing
+    return
+end
+
+# Archive files are named `<device>-<OS build>.bin`; the device name is sanitized to a
+# conservative character set so that the build suffix can be recognized unambiguously.
+const binary_archive_ext = ".bin"
```
christiangnrd: Apple's [example](https://developer.apple.com/documentation/metal/creating-binary-archives-from-device-built-pipeline-state-objects?language=objc) seems to use `.binary.metallib`. Should we also use that?
```suggestion
const binary_archive_ext = ".binary.metallib"
```

maleadt:
With all these improvements, cached workloads are pretty quick:

```julia
module Foo

using Metal
using Metal: @setup_workload, @compile_workload

function vadd_kernel(a, b, c)
    i = thread_position_in_grid().x
    c[i] = a[i] + b[i]
    return
end

function vadd(d_a, d_b, d_c)
    dims = size(d_c)
    len = prod(dims)
    @metal threads=len vadd_kernel(d_a, d_b, d_c)
    Array(d_c)
end

@setup_workload begin
    dims = (3,4)
    a = round.(rand(Float32, dims) * 100)
    b = round.(rand(Float32, dims) * 100)
    c = similar(a)

    d_a = MtlArray(a)
    d_b = MtlArray(b)
    d_c = MtlArray(c)

    @compile_workload vadd(d_a, d_b, d_c)
end

end
```

```julia-repl
julia> using Metal, Foo

julia> dims = (3,4);
julia> a = round.(rand(Float32, dims) * 100);
julia> b = round.(rand(Float32, dims) * 100);
julia> c = similar(a);

julia> d_a = MtlArray(a);
julia> d_b = MtlArray(b);
julia> d_c = MtlArray(c);

julia> @time Foo.vadd(d_a, d_b, d_c)
  0.011843 seconds (1.27 k allocations: 200.312 KiB, 27.73% compilation time)

julia> @time Foo.vadd(d_a, d_b, d_c)
  0.001558 seconds (63 allocations: 2.109 KiB)
```

PR #944, opened 2026-08-29
Fix loading on Julia master

christiangnrd:
Inspired by https://github.com/JuliaGPU/OpenCL.jl/pull/477

### DIFF
--- a/src/device/quirks.jl
+++ b/src/device/quirks.jl
@@ -1,4 +1,8 @@
 # math.jl
+@static if isdefined(Base.Math, :throw_finite_domainerror)
+    @device_override @noinline Base.Math.throw_finite_domainerror(f::Symbol, x) =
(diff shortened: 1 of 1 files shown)


src/device/quirks.jl
```diff
@@ -1,4 +1,8 @@
 # math.jl
+@static if isdefined(Base.Math, :throw_finite_domainerror)
+    @device_override @noinline Base.Math.throw_finite_domainerror(f::Symbol, x) =
+        @gputhrow "DomainError" "(x) is only defined for finite x."
```
yuyichao, line 4: Note that in the OpenCL version I passed the function name to the error printing as well. I’m not sure if gputhrow supports it here but if now you might want to change the error message a little bit. Sth like function is only defined for … might be better.

christiangnrd: Good point thank you!

PR #949, opened 2026-09-03
Use `mul!` in flopscomp

christiangnrd:
Follow up to https://github.com/JuliaGPU/Metal.jl/pull/948. Seems like buildkite CI was still on 1.13 rc3 so this was missed. I ran the testsuite locally and they passed


maleadt:
Why was this calling into the internal function anyway?

christiangnrd:
When I combined these to use the scopedvalues algorithm selection I based it on the GPUArrays.generic_matmatmul version and I didn't think to switch to `mul!`

maleadt (approved):

PR #953, opened 2026-09-08
Support `cbrt(::Float32)`

### LINKED ISSUE #952: `cbrt(::Float32)` requires Float64
```julia-repl
julia> using Metal

julia> cbrt.(Metal.ones(Float32, 1))
ERROR: InvalidIRError: compiling MethodInstance for (::Metal.var"#broadcast_linear#_copyto!##1")(::MtlDeviceVector{Float32, 1}, ::Base.Broadcast.Broadcasted{Metal.MtlArrayStyle{…}, Tuple{…}, typeof(cbrt), Tuple{…}}) resulted in invalid LLVM IR
Reason: unsupported use of double value
Stacktrace:
 [1] Float64
   @ ./float.jl:345
 [2] _improve_cbrt
   @ ./special/cbrt.jl:87
 [3] cbrt
   @ ./special/cbrt.jl:147
 [4] _broadcast_getindex_evalf
   @ ./broadcast.jl:699
 [5] _broadcast_getindex
   @ ./broadcast.jl:672
 [6] _getindex
   @ ./broadcast.jl:620
 [7] getindex
   @ ./broadcast.jl:616
 [8] broadcast_linear
   @ ~/.julia/packages/Metal/VXhss/src/broadcast.jl:97
[...]

julia> Metal.versioninfo()
macOS 15.5.0, Darwin 24.5.0

Toolchain:
- Julia: 1.12.7
- LLVM: 18.1.7
- Metal: 3.2 (MSL), 2.7 (AIR), 1.2.8 (metallib)

Julia packages:
- Metal.jl: 1.10.3
- GPUArrays: 11.5.14
- GPUCompiler: 2.5.4
- KernelAbstractions: 0.9.42
- ObjectiveC: 6.0.1
- LLVM: 9.13.1
- LLVMDowngrader_jll: 0.9.1+0

1 device:
- Apple M1 (8 GPU cores, 1.375 MiB allocated; Apple7, Metal3 family)
```

maleadt:
Base refines Float32 cube roots in Float64, which Metal does not support. Use integer argument reduction, a polynomial estimate, and a compensated Newton correction to keep all arithmetic in Float32 while preserving subnormal inputs.

Test special values, exponent and subnormal boundaries, and accuracy against CPU references for Float32 and Float16.

Fixes #952.

### DIFF
--- a/docs/src/faq/faq.md
+++ b/docs/src/faq/faq.md
@@ -5,3 +5,16 @@
 Most likely. Any help on designing or implementing high-level wrappers for MSL's low-level functionality
 is greatly appreciated, so please consider [contributing](contributing.md) your uses of these APIs on the
 respective repositories.
+
+## Why do Float32 subnormals become zero on the GPU?
+
+On Apple GPUs, `Float32` arithmetic flushes subnormal operands and results to zero.
+Comparisons also treat subnormal operands as zero, so `iszero(x)` and `x == 0` can be true
+for a nonzero value stored in memory. Multiplying a subnormal input by a power of two
+therefore does not recover its value.
+
+The [Metal Shading Language specification](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf)
+(section 8.1) permits flushing subnormals to zero, including for `Float16`; code should not
+assume that disabling fast math preserves them. Algorithms that need to handle subnormal
+inputs, such as `cbrt`, can normalize their bit patterns using integer operations before
+performing floating-point arithmetic.
--- a/src/device/intrinsics/math.jl
+++ b/src/device/intrinsics/math.jl
@@ -456,3 +456,56 @@ end
 
     return r
 end
+
+### Cube root
+
+# Base refines cbrt(::Float32) in Float64, which Metal does not support.
+# Reduce the argument to [1, 8), approximate with a polynomial, then apply a
+# compensated Newton step. Integer reduction preserves subnormal inputs despite FTZ.
+# Exhaustive testing of positive finite Float32 inputs on an M3 Pro measured a
+# maximum error of 0.50003 ulp; this is not a correctly rounded implementation.
+@device_override function Base.cbrt(x::Float32)
+    u = reinterpret(UInt32, x) & 0x7fffffff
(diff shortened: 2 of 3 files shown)


maleadt:
@oscardssmith Does this look reasonable?

maleadt:
Compared to the `copysign(abs(x)^(1f0/3f0), x)` from https://github.com/CliMA/Oceananigans.jl/pull/5940:

| | `pow` workaround | this PR |
|---|---|---|
| 2^24 random normal inputs: correctly rounded / ≤ 1 ulp / ≤ 4 ulp | 5% / 10% / 39% | 100% / 100% / 100% |
| same: max / mean error | 15.9 / 5.4 ulp | 0.50 / 0.25 ulp |
| 2^24 inputs in `[0.25, 16)`: correctly rounded, max error | 65%, 1.6 ulp | 100%, 0.50 ulp |
| subnormal inputs | all return 0 | exact except 84 off by 1 ulp |
| exhaustive over all Float32 | 95% not correctly rounded | 0.001% off by 1 ulp, max 0.50003 ulp |
| dependent chain, 64 × 2^22 evaluations (ALU bound) | 13.4 ms | 7.3 ms |
| in-place broadcast, 2^26 elements (memory bound) | 6.1 ms | 4.1 ms (same as a plain `fma`) |

> The `pow` error grows with `|log2(x)|`, since `1f0/3f0` is not one third and `air.pow` is itself relaxed precision: it is a 1–2 ulp function near 1 and a 16 ulp function at the ends of the range. Adding a Newton step on top of `pow` does not fix it either, because for `|x| < 2^-104` the residual `t^3 - x` is subnormal and gets flushed to zero. This implementation avoids that by reducing the argument to `[1, 8)` with integer operations and doing the Newton step there.
>
> So the one-liner is about 2× slower and loses 3–4 bits over most of the range, on top of zeroing subnormals. Fine for a turbulent length scale, but not as a `Base.cbrt` override.

oscardssmith:
How is the speed of this? would this be a good method for Base?

maleadt:
It's about 20% slower than Base, so no. On CPUs, that is, on GPUs it's likely faster due to fewer `div` instructions.

PR #956, opened 2026-09-09
Adapt to the library-based LLVMDowngrader_jll 0.10

maleadt:
JuliaPackaging/Yggdrasil#14727 replaced the `llc`, `lld` and `llvm-downgrade` executables in the GPU LLVM JLLs with shared libraries exposing a small C API, moving the back-ends to LLVM 23 at the same time. JuliaGPU/GPUCompiler.jl#930 makes GPUCompiler call those libraries in-process, and bumps its compat to the new JLL majors. Since this package pins the same JLL, it needs its compat bumped in lockstep, which is what this PR does.

Changes: `LLVMDowngrader_jll` compat → 0.10 and `GPUCompiler` compat → 2.7. Metal.jl only imports the JLL so that GPUCompiler can use it. The downgrader no longer ships the legacy `llvm-dis-14`, so `code_native`/`code_air` output is now disassembled by the in-process LLVM: it reads as that LLVM's dialect (opaque pointers) rather than LLVM 14's. The bitcode that goes into the metallib is unchanged. The simdgroup intrinsic ABI tests matched the typed-pointer spelling and now accept either.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -41,6 +41,9 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 [extensions]
 SpecialFunctionsExt = "SpecialFunctions"
 
+[sources]
+GPUCompiler = {url = "https://github.com/JuliaGPU/GPUCompiler.jl", rev = "tb/jll-libraries"}
+
 [compat]
 AbstractFFTs = "1"
 Adapt = "4.6.1"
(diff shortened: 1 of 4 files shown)


christiangnrd:
Is this in a state where we could sneak it into Metal.jl 1.11?

maleadt:
> Is this in a state where we could sneak it into Metal.jl 1.11?

Sure, I think so.

PR #966, opened 2026-09-11
Try macOS 27 in CI

christiangnrd:


maleadt:
Is it released already?

christiangnrd:
macOS 27 is out in 3 days. I just happened to notice that GitHub had an xcode-27 runner that seems to have been on the betas (and now on RC)

I don’t think we should merge this until the actual macos-27 runner is announced.

christiangnrd:
https://github.com/JuliaGPU/Metal.jl/issues/968 made me give the failure a second look and I don't think it's related to that issue but I think the atomic ordering stuff might not be supported on paravirtual devices

